PR #258, opened 2014-02-11
Functions of Database and Statement should be domain-aware

FrozenCow:
This is based on a earlier pull request that became stale a year ago: https://github.com/mapbox/node-sqlite3/pull/81

It is about NodeJS domains: http://nodejs.org/api/domain.html

At the moment callbacks from Database and Statement are not domain-aware. That means that when a call was made from within a domain, the callback of that call won't be in that same domain. Exceptions that are thrown inside such a callback will be caught by process.on('uncaughtException') instead of domain.on('error'). This can be confusing for developers that make use of domains for exceptions or logging.

I used the commits from @chrisdickinson as a base and converted the test to Mocha.
I've also incorporated the comments from the mentioned pull request, which basically said to make use of EventEmitter to set domain, instead of manually setting the domain. To do this I had to call back to JS from the native Database/Statement C++ constructors.
Database#exec is not implemented using Statement, so the callback for that function needed to be wrapped in order to pass the domain.

Side note:
This is a step in the right direction I think, but this doesn't make everything of node-sqlite3 domain-aware. The calls in Statement are purely native and I wasn't sure how to store and bind the domain to the callback. This is however quite easy if these functions were wrapped in a Javascript. Database#exec is an example for that.
That said, problems/confusion will only arise once a statement is created in one domain (A), but the statement is used in another (B). You would think callbacks will be in domain B, but they will be in domain A with these commits.

### DIFF
--- a/lib/sqlite3.js
+++ b/lib/sqlite3.js
@@ -22,7 +22,7 @@ function inherits(target, source) {
 
 sqlite3.cached = {
     Database: function(file, a, b) {
-        EventEmitter.call(this)
+        EventEmitter.call(this);
 
         if (file === '' || file === ':memory:') {
             // Don't cache special databases.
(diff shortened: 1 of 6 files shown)


hillmanov:
Looks like this hasn't been discussed in a while - is this still something that is going to be considered? If not, is there a recommended approach to handling errors that are thrown that are non-catchable with try/catch or with domains?

springmeyer:
> Looks like this hasn't been discussed in a while - is this still something that is going to be considered?

@hillmanov I don't use domains / have much experience with them, so I've been hesitant to accept this but I'm open to it still.

> If not, is there a recommended approach to handling errors that are thrown that are non-catchable with try/catch or with domains?

Can you explain what kind of errors these are?

FrozenCow:
Domains shouldn't be used to actually catch errors, just to log/know in what context the errors happened (after handling the error you should kill the process anyway). It is also now deprecated in the latest Node versions: https://nodejs.org/api/domain.html See the discussion here: https://github.com/nodejs/node/issues/66

An alternative would be https://github.com/btford/zone.js/, but I haven't given it a try yet.

wmertens:
in Node v9, domains are deprecated…

Issue #353, opened 2014-10-14
Coredump when using prepared statements


Kasalehlia:
When using the Function.prototype.call or Function.prototype.apply functions on the run/get/all/each function of a preapred statement, not submitting the statement as `this` argument results in the message shown below:

```
node: /home/travis/.node-gyp/0.10.29/src/node_object_wrap.h:61: static T* node::ObjectWrap::Unwrap(v8::Handle<v8::Object>) [with T = node_sqlite3::Statement]: Assertion `handle->InternalFieldCount() > 0' failed.
```

I am using sqlite3@3.0.2 and node v0.10.32.
Example code to trigger this issue:

```
var sqlite3 = require('sqlite3');
db = new sqlite3.Database(':memory:', function () {
    statement = db.prepare('CREATE TABLE a (id TEXT);');
    statement.run.call(null);
});
```

Using the same code on the run function of the database itself triggers an exception: `TypeError: Database object expected` which I suppose is the correct behaviour in that case.

daniellockyer:
This no longer crashes and instead throws an `Illegal invocation error`

PR #360, opened 2014-10-31
added 'this' check for prepared statements (#353)

Kasalehlia:
To address #353, I included a HasInstance check in every function using args.This()


kkoopa:
Could use `NanHasInstance` to avoid the conversion for Node 0.10 and below.

springmeyer:
@Kasalehlia- would you mind rebasing against latest master? Then if this stays passing I'll merge. Thanks!

tmcw:
closing, stale

PR #370, opened 2014-11-25
fix make binding.gyp compatible with node-gyp

bwin:
Instead of just node-pre-gyp.

The binding.gyp in current master isn't compatible with node-gyp without manual editing. (AFAIK)

BTW I'm interested in this, because this allows for automated building for atom-shell with node-gyp.
(typo in the commit msg where it says nw-gyp)

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -2,11 +2,12 @@
   "includes": [ "deps/common-sqlite.gypi" ],
   "variables": {
       "sqlite%":"internal",
-      "sqlite_libname%":"sqlite3"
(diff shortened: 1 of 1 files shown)


springmeyer:
> But this doesn't break anything

It does: it breaks the require (https://github.com/mapbox/node-sqlite3/blob/d48947d71d450c4763d887ac3c4d39346bb0cc97/lib/sqlite3.js#L3) which looks inside `./lib/binding/{node_abi}-{platform}-{arch}` not `./lib/binding`.

Issue #445, opened 2015-05-05
Failure to build 3.0.6 on ARM


Page-:
I'm getting this issue when trying to build 3.0.6 on nodejs 0.10.38 for ARM (using qemu): `/usr/bin/ld: Release/obj.target/node_sqlite3.node: No symbol version section for versioned symbol `memcpy@GLIBC_2.2.5'`

Full log:
```
npm install sqlite3
-
> sqlite3@3.0.6 install /usr/src/node_modules/sqlite3
> node-pre-gyp install --fallback-to-build

make: Entering directory '/usr/src/node_modules/sqlite3/build'
  ACTION deps_sqlite3_gyp_action_before_build_target_unpack_sqlite_dep Release/obj/gen/sqlite-autoconf-3080900/sqlite3.c
  TOUCH Release/obj.target/deps/action_before_build.stamp
  CC(target) Release/obj.target/sqlite3/gen/sqlite-autoconf-3080900/sqlite3.o
  AR(target) Release/obj.target/deps/sqlite3.a
  COPY Release/sqlite3.a
  CXX(target) Release/obj.target/node_sqlite3/src/database.o
../src/database.cc: In static member function 'static void node_sqlite3::Database::Work_BeginOpen(node_sqlite3::Database::Baton*)':
../src/database.cc:145:9: warning: unused variable 'status' [-Wunused-variable]
     int status = uv_queue_work(uv_default_loop(),
         ^
../src/database.cc: In static member function 'static void node_sqlite3::Database::Work_BeginClose(node_sqlite3::Database::Baton*)':
../src/database.cc:230:9: warning: unused variable 'status' [-Wunused-variable]
     int status = uv_queue_work(uv_default_loop(),
         ^
../src/database.cc: In static member function 'static void node_sqlite3::Database::Work_BeginExec(node_sqlite3::Database::Baton*)':
../src/database.cc:509:9: warning: unused variable 'status' [-Wunused-variable]
     int status = uv_queue_work(uv_default_loop(),
         ^
../src/database.cc: In static member function 'static void node_sqlite3::Database::Work_BeginLoadExtension(node_sqlite3::Database::Baton*)':
../src/database.cc:610:9: warning: unused variable 'status' [-Wunused-variable]
     int status = uv_queue_work(uv_default_loop(),
         ^
  CXX(target) Release/obj.target/node_sqlite3/src/node_sqlite3.o
  CXX(target) Release/obj.target/node_sqlite3/src/statement.o
[...]

PR #446, opened 2015-05-05
Fix non-x86 build failure.

bnoordhuis:
The inclusion of gcc-preinclude.h breaks the build on ARM because of
the missing .symver symbol.  Only define it on x86.

Closes out mapbox/node-sqlite3#445.


bnoordhuis:
@springmeyer Updated, PTAL.

springmeyer:
thank you @bnoordhuis - will merge and make another quick release: v3.0.7

PR #447, opened 2015-05-07
Fix non-Linux build failure.

bnoordhuis:
gcc-preinclude.h contains a fix for old Linux systems but it's included
unconditionally on other platforms as well.

I could update the GYP files to exclude it on other platforms but it
seems easier to add a `defined(__linux__)` guard to the header file.

Let this be the last fix.

Refs #445.  R=@springmeyer?


springmeyer:
Ah, this is needed I presume for Sun/SmartOS?

bnoordhuis:
@basarevych reported a FreeBSD build failure in #445 but it should help on Solaris as well.

springmeyer:
thanks for the details and the fix :)

Issue #376, opened 2014-12-10
Feature request:  emit declared field types (sqlite3_column_decltype()) from query method results


rogierschouten:
I know, sqlite3 does not have static field typing. Even so, columns in sqlite3 have a **declared** type which can be retrieved with the sqlite3_column_decltype() function.  

I'm making a database abstraction layer, and I'd like to include an adapter for sqlite3. The only piece of the puzzle that is missing for my use case is that I need to get the query result declared column types even if no rows are returned. Would it be possible to add this information to the callback of Database#all(), and an extra callback to Database#each(), which is called before the row/complete callbacks?

PR #451, opened 2015-05-14
Added declared type values as 3rd argument to callbacks

wbyoung:
This fully addresses #376.

I think this is a really valuable feature that many of the other database bindings provide. This should be a relatively straightforward change of simply passing an object containing the column names as properties and their declared types as the values.

The only note on backwards compatibility is that we're now passing 3 arguments to callbacks. If someone used an assertion/guard for `arguments.length === 2`, then their code would break. I doubt a large percentage of code does this, and the documentation doesn't really indicate that more arguments would never be passed.

### DIFF
--- a/src/statement.cc
+++ b/src/statement.cc
@@ -380,6 +380,15 @@ void Statement::Work_Get(uv_work_t* req) {
         sqlite3_mutex_enter(mtx);
 
         if (stmt->Bind(baton->parameters)) {
+            int cols = sqlite3_column_count(stmt->_handle);
+            for (int i = 0; i < cols; i++) {
+                const char *name = sqlite3_column_name(stmt->_handle, i);
+                const char *type = sqlite3_column_decltype(stmt->_handle, i);
+                if (type) {
+                    baton->types[name] = type;
+                }
+            }
+
             stmt->status = sqlite3_step(stmt->_handle);
 
             if (!(stmt->status == SQLITE_ROW || stmt->status == SQLITE_DONE)) {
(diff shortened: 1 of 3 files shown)


emilioplatzer:
+1

kroggen:
Hi @wyoung!

Thank you for your work!

I have a modest suggestion:

I noticed that some of your code is reused. So I was wondering if they could be in functions.

This block here appears 3 times and could be in a function receiving (stmt->_handle, baton or baton->types):

        int cols = sqlite3_column_count(stmt->_handle);
        for (int i = 0; i < cols; i++) {
            const char *name = sqlite3_column_name(stmt->_handle, i);
            const char *type = sqlite3_column_decltype(stmt->_handle, i);
            if (type) {
                baton->types[name] = type;
            }
        }

And this one here also appears 3 times and could be in another function that returns the decltypes map already filled:

            Local<Object> decltypes(NanNew<Object>());
            Decltypes::const_iterator type_it = baton->types.begin();
            Decltypes::const_iterator type_end = baton->types.end();
            for (; type_it != type_end; ++type_it) {
                Local<String> name = NanNew<String>(type_it->first.c_str());
                Local<String> type = NanNew<String>(type_it->second.c_str());
                decltypes->Set(name, type);
            }

kroggen:
+1

Hi @wbyoung!

I also need something like this.

Thank you for your work!

I have a modest suggestion:

I noticed that some of your code is reused. So I was wondering if they could be in functions.

This block here appears 3 times and could be in a function receiving (stmt->_handle, baton or baton->types):

```c++
        int cols = sqlite3_column_count(stmt->_handle);
        for (int i = 0; i < cols; i++) {
            const char *name = sqlite3_column_name(stmt->_handle, i);
            const char *type = sqlite3_column_decltype(stmt->_handle, i);
            if (type) {
                baton->types[name] = type;
            }
        }
```
And this one here also appears 3 times and could be in another function that returns the decltypes map already filled:

```c++
            Local<Object> decltypes(NanNew<Object>());
            Decltypes::const_iterator type_it = baton->types.begin();
            Decltypes::const_iterator type_end = baton->types.end();
            for (; type_it != type_end; ++type_it) {
                Local<String> name = NanNew<String>(type_it->first.c_str());
                Local<String> type = NanNew<String>(type_it->second.c_str());
                decltypes->Set(name, type);
            }
```

I am not a C++ developer so I don't know if I can do this.

Or whether it'is a good thing to use functions in this case...

wbyoung:
@kroggen thanks for the suggestion. If the project owners want that done, I'll try to find the time to update the PR.

bedwardly-down:
@wbyoung, since this PR is almost four years old, is there any valid reason to keep it open? If so, are you able to update it to a state where it can be tested and possibly merged if needed? If not, could you please close this PR out since it's no longer being maintained and has no activity on it? Thank you.

daniellockyer:
The PR sounds useful but it uses NAN and we've since switched to N-API. If it could be reworked to N-API, I think it's worth looking again 🙂

PR #461, opened 2015-06-14
unpacked iojs tar file ends with -ia32 and not -x86

ralphtheninja:
Noticed this while playing around with your build scripts. Using them as inspiration for `leveldown`.


ralphtheninja:
*bump* :)

springmeyer:
Thanks for noticing this. However I think it should be fixed upstream instead: https://github.com/nodejs/build/issues/45

ralphtheninja:
:+1: You could also switch to `n` if you like, after https://github.com/tj/n/pull/282 has been merged, or just use my fork of it. Then you can install 32 bit binaries on 64 systems and don't have to use `curl` explicitly.

ralphtheninja:
See https://github.com/Level/leveldown/commit/7b21861b34a1e897f58afff5a9c65d6eff807db2 for more details. I based it on your current script :)

Issue #472, opened 2015-06-29
Failed to build manually


coolaj86:
Solution
=======

* [Building SQLCipher for node.js and Raspberry Pi](https://coolaj86.com/articles/building-sqlcipher-for-node-js-on-raspberry-pi-2/)
* [Building SQLCipher for OS X and node.js](https://coolaj86.com/articles/building-sqlcipher-on-os-x/)

Problem
=======

When I try to follow the directions in the README this is what I get:

```
export LDFLAGS="-L/usr/local/lib"
export CPPFLAGS="-I/usr/local/include"
npm install sqlite3 --build-from-source --sqlite_libname=sqlcipher --sqlite=/usr/local



\
> sqlite3@3.0.8 install /home/pi/node_modules/sqlite3
> node-pre-gyp install --fallback-to-build

make: Entering directory '/home/pi/node_modules/sqlite3/build'
  CXX(target) Release/obj.target/node_sqlite3/src/database.o
In file included from ../src/database.cc:5:0:
../src/database.h:10:21: fatal error: sqlite3.h: No such file or directory
compilation terminated.
node_sqlite3.target.mk:93: recipe for target 'Release/obj.target/node_sqlite3/src/database.o' failed
make: *** [Release/obj.target/node_sqlite3/src/database.o] Error 1
make: Leaving directory '/home/pi/node_modules/sqlite3/build'
```

Yet we can see that `sqlite3.h` most certainly does exist:

```
ls /usr/local/include/sqlcipher/sqlite3.h
```

This is a leftover from https://github.com/mapbox/node-sqlite3/issues/413

springmeyer:
Can you pass `--verbose` as well so we can see the compile args. My guess is that `-I/usr/local/include` is not ending up being passed to the compiler. You could try `export CXXFLAGS="-I/usr/local/include"` instead of CPPFLAGS.

springmeyer:
Also try:  `export CPPFLAGS="-I/usr/local/include/sqlcipher/"`

coolaj86:
```
export LDFLAGS="-L/usr/local/lib"
export CPPFLAGS="-I/usr/local/include"
export CXXFLAGS="-I/usr/local/include"
npm install sqlite3 --build-from-source --sqlite_libname=sqlcipher --sqlite=/usr/local --verbose
```


```
npm info it worked if it ends with ok
npm verb cli [ '/usr/local/bin/iojs',
npm verb cli   '/usr/local/bin/npm',
[...]

Issue #414, opened 2015-02-22
include_dirs not being populated by ./configure or node-gyp configure


coolaj86:
# Solution
- [Building SQLCipher for node.js and Raspberry Pi](https://coolaj86.com/articles/building-sqlcipher-for-node-js-on-raspberry-pi-2/)
- [Building SQLCipher for OS X and node.js](https://coolaj86.com/articles/building-sqlcipher-on-os-x/)
# Problem

**Download**: https://dropsha.re/files/smart-impala-96/sqlcipher-arm6l-node-v0.10.tar.gz

In relation to #412 and #413, there's a common thread: It appears that "include_dirs" isn't being populated.

Not only that, but it appears to need to be populated with the _incorrect_ value to work correctly.

I've repeated this process a few times just to make sure, so I'm fairly confident I've caught all of the relevant details:
### 1. clone and configure

``` bash
npm install node-gyp -g

git clone git@github.com:mapbox/node-sqlite3.git
pushd node-sqlite3

npm install nan
npm install node-pre-gyp
rsync -av node_modules/ ../node_modules/

export USR_LOCAL=/usr/local
export LDFLAGS="-L${USR_LOCAL}/lib"
export CPPFLAGS="-I${USR_LOCAL}/include"

# make style
# removes node_modules... why?
make clean
# and just to be safe
node-gyp clean
./configure --build-from-source --sqlite_libname=sqlcipher --sqlite=/usr/local --name=sqlite3 --configuration=Release --module_name=node_sqlite3 --module_path=$(pwd)/lib/binding/node-v10-linux-arm
```

**NOTE**: I had to manually specify `module_name` and `module_path`. Even still, **module_path is ignored**.
### 2. Edit build/config.gyp manually

`build/config.gypi` should contain

```
"include_dirs": ["/usr/local/include/sqlcipher"],
```

**Shouldn't this be "/usr/local/include"?** That's what `binding.gyp` would suggest.

I also tried with `["/usr/local/include/sqlcipher", "/usr/local/include", "/usr/local"]`, but the above seemed to be the minimum requirement after trying them one at a time (so it could be that one got partially added each time).
### 3. Build

this will only take a minute since the bulk of the work was done when you build sqlcipher

``` bash
rsync -av ../node_modules/nan node_modules/
rsync -av ../node_modules/node-pre-gyp node_modules/
rsync -av ../node_modules/.bin node_modules/
make

[...]

Issue #413, opened 2015-02-22
Failed to build manually - can't find sqlite3.h (even with --sqlite=/usr/local)


coolaj86:
**Update**: Workaround available at https://github.com/mapbox/node-sqlite3/issues/414

Here's what I've done:

```bash
npm install node-gyp -g
npm install -g nan

git clone git@github.com:mapbox/node-sqlite3.git
pushd node-sqlite3

npm install nan

node-gyp clean
node-gyp configure --sqlite=/usr/local --sqlite_libname=sqlcipher --module_name=node_sqlite3 --module_path=$(pwd)/lib/binding/node-v10-linux-arm
rsync -av /usr/local/include/sqlcipher/sqlite3* src/

node-gyp build
```

Not sure what I'm doing wrong here, but I just can't get it to find sqlite3.h

springmeyer:
calling node-gyp directly is not supported as mentioned at https://github.com/mapbox/node-sqlite3/issues/416#issuecomment-75415682

Issue #412, opened 2015-02-21
Failed to build for sqlcipher on RaspberryPi with node 10.36


coolaj86:
**Update**: Workaround available at https://github.com/mapbox/node-sqlite3/issues/414
**Download**: https://dropsha.re/files/smart-impala-96/sqlcipher-arm6l-node-v0.10.tar.gz

As per https://gist.github.com/coolaj86/4904b0b471be660c6881#file-raspberrypi-md

Here's what I did:

On the Pi this will take a half hour or so

```bash
sudo apt-get install -y build-essential

apt-cache search openssl
sudo apt-get install -y openssl libssl-dev

git clone https://github.com/sqlcipher/sqlcipher.git
pushd sqlcipher

./configure --help
./configure CFLAGS="-DSQLITE_HAS_CODEC -DSQLITE_ENABLE_FTS3 -lcrypto"

make

sudo chown -R $(whoami):$(whoami) /usr/local

make install
```

And the npm install process will take about the same time, maybe a little bit quicker

```bash
export USR_LOCAL=/usr/local
export LDFLAGS="-L${USR_LOCAL}/lib"
export CPPFLAGS="-I${USR_LOCAL}/include"
npm install sqlite3 --build-from-source --sqlite_libname=sqlcipher --sqlite=${USR_LOCAL}
```

And it failed with this

```
gyp ERR! cwd /srv/pi/node_modules/sqlite3
gyp ERR! node -v v0.10.36
gyp ERR! node-gyp -v v1.0.1
gyp ERR! not ok
node-pre-gyp ERR! build error
[...]

PR #473, opened 2015-06-29
correct instructions for building against sqlcipher

coolaj86:
add common instructions for linux as well as link to sqlcipher build instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,15 +136,38 @@ Visit the “[Using Node modules](https://github.com/rogerwang/node-webkit/wiki/
 
 ## Building for sqlcipher
 
+For instructions for building sqlcipher see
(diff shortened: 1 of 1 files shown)


coolaj86:
Related to
* https://github.com/mapbox/node-sqlite3/issues/472
* https://github.com/mapbox/node-sqlite3/issues/414
* https://github.com/mapbox/node-sqlite3/issues/413
* https://github.com/mapbox/node-sqlite3/issues/412

springmeyer:
I don't like the hardcoding of paths in `-I/usr/local/include -I/usr/local/include/sqlcipher"` for the docs. That might work on one system but not the next. Can you think of a way to make this more universal and less hardcoded?

coolaj86:
It's clearly labeled *for most linuxes*. There's no sense in trying to get fancy.

If someone is doing something weird, they're going to know. For people that are just following average-joe instructions, it's always going to be `/usr/local/`.

PR #486, opened 2015-07-30
Upgrade to Sqlite 3.8.11.1

bengotow:
Sqlite 3.8.11.1 was released this morning with some bug fixes, and version 3.8.11 also saw micro-optimizations that make it considerably faster: "22.3% more work for the same number of CPU cycles relative to the previous release. SQLite now runs twice as fast as version 3.8.0 and three times as fast as version 3.3.9."

I upgraded the sqlite binary, bumped the version number, and checked that the npm install and test scripts still run successfully!


bengotow:
(looking at the failing Appveyor check, I'm not sure it's related to this change)

springmeyer:
Thanks!

springmeyer:
btw appveyor fails will be fixed by https://github.com/mapbox/node-sqlite3/pull/488

Issue #459, opened 2015-06-11
npm install sqlite3 --build-from-source (fails)


fractalf:
I've had heaps of problems with sqlite3 earlier, but mostly managed to figure it out by reading the errors carefully. This time however I'm quite stuck. Any ideas?

```
FROM alpine:edge

RUN echo "http://nl.alpinelinux.org/alpine/edge/main" > /etc/apk/repositories && \
    echo "http://nl.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories && \
    apk add -U iojs wget ca-certificates python make gcc libc-dev g++ && \
    apk upgrade && \
    npm install sqlite3 --build-from-source
```

this will give error

```
...
../src/statement.cc: In static member function 'static void node_sqlite3::Statement::Work_BeginReset(node_sqlite3::Statement::Baton*)':
../src/macros.h:122:9: warning: unused variable 'status' [-Wunused-variable]
     int status = uv_queue_work(uv_default_loop(),                              \
         ^
../src/statement.cc:725:5: note: in expansion of macro 'STATEMENT_BEGIN'
     STATEMENT_BEGIN(Reset);
     ^
  SOLINK_MODULE(target) Release/obj.target/node_sqlite3.node
/usr/lib/gcc/x86_64-alpine-linux-musl/5.1.0/../../../../x86_64-alpine-linux-musl/bin/ld: Release/obj.target/node_sqlite3.node: No symbol version section for versioned symbol `memcpy@GLIBC_2.2.5'
/usr/lib/gcc/x86_64-alpine-linux-musl/5.1.0/../../../../x86_64-alpine-linux-musl/bin/ld: final link failed: Nonrepresentable section on output
collect2: error: ld returned 1 exit status
node_sqlite3.target.mk:130: recipe for target 'Release/obj.target/node_sqlite3.node' failed
make: *** [Release/obj.target/node_sqlite3.node] Error 1
make: Leaving directory '/node_modules/sqlite3/build'
gyp ERR! build error
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:269:23)
gyp ERR! stack     at emitTwo (events.js:87:13)
gyp ERR! stack     at ChildProcess.emit (events.js:172:7)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:199:12)
[...]

PR #489, opened 2015-08-06
Remove memcpy@@GLIBC_2.14 workaround.

bnoordhuis:
There is only one place in the code base that uses memcpy().  Simply
rewrite it to not depend on mempcy() and drop the symver hack.

Fixes: https://github.com/mapbox/node-sqlite3/issues/459

R=@springmeyer

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -31,7 +31,6 @@
         }
         ]
       ],
-      "cflags": [ "-include ../src/gcc-preinclude.h" ],
       "sources": [
         "src/database.cc",
         "src/node_sqlite3.cc",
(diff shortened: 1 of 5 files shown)


springmeyer:
@bnoordhuis - thank you. This drop of `memcpy` in the `node-sqlite3` code looks good. However the bundled libsqlite3 dep still uses `memcpy` extensively. So I'm assuming that the statically linked libsqlite3 code will still be a problem?

bnoordhuis:
`git grep memcpy` didn't turn up anything besides that one reference in statement.h but I see now that sqlite3 is stored as a tarball... okay, closing, this isn't going to work.

PR #505, opened 2015-09-09
Add support for travis docker + node 4.x

springmeyer:


springmeyer:
one glitch with this will be addressed separately: #506

Issue #503, opened 2015-09-09
stmt.finalize(callback) can't promise the sequence


march1993:
codes like:
```javascript
db.serialize();
var result = []
stmt.prepare(...);
stmt.each(parameter, function(err, row) {
    result.push(row['some_column']);
});
stmt.finalize(function () {
    console.log(JSON.stringify(result));
}
```
but result is missing some rows

it behaves right on Windows7 x64
but behaves not right on Centos 4.3 x64

springmeyer:
@march1993 - can you provide a pull request with a unit test - then the test will run on ubuntu precise and we can see if it works?

march1993:
let me have a try

march1993:
see https://github.com/mapbox/node-sqlite3/pull/507

LivorMortis:
@march1993: I came across this issue when I ran into the same problem, until I realized that finalize doesn't guarantee that the query has been executed in its callback.  Instead, you should use the _complete_ argument of **Statement#each([param, ...], [callback], [complete])**

So rewriting your example code to the following should work:
```node
db.serialize();
var result = []
stmt.prepare(...);
stmt.each(parameter, function(err, row) {
    result.push(row['some_column']);
}, function(err, numRows) {
    console.log(JSON.stringify(result));
});
stmt.finalize();
```

tmcw:
Yep, workaround is ^, use the second argument of .each

PR #507, opened 2015-09-10
mapbox/node-sqlite3/issues/503 test

march1993:
bug


tmcw:
Fix is explained in the bug report ticket - use the second argument of .each: https://github.com/mapbox/node-sqlite3/issues/503#issuecomment-219820887

PR #518, opened 2015-09-26
Add support for sqlite3_interrupt

briangreenery:
This change adds support for [sqlite3_interrupt](https://www.sqlite.org/c3ref/interrupt.html). This makes it possible to interrupt a long running query.

I called the function synchronously instead of going through `db->Schedule` because if it was scheduled it wouldn't be possible to interrupt `exec`. This is because while `exec` is running the work queue is not processed. Also, the `sqlite3_interrupt` function doesn't do any I/O or acquire any locks, so it won't block the javascript thread.

The downside of calling `sqlite3_interrupt` synchronously means that it has to check whether the database is open, and has to check that `sqlite3_close` is not about to be run by another thread. Basically, if `sqlite3_interrupt` is called on an already closed database, then it could cause memory corruption.

To avoid this, I added a `closing` variable that's set right before sending a `close` to a worker thread. The `open` variable isn't sufficient because it's only set to `false` after `sqlite3_close` has been called.

### DIFF
--- a/src/database.cc
+++ b/src/database.cc
@@ -23,6 +23,7 @@ NAN_MODULE_INIT(Database::Init) {
     Nan::SetPrototypeMethod(t, "serialize", Serialize);
     Nan::SetPrototypeMethod(t, "parallelize", Parallelize);
     Nan::SetPrototypeMethod(t, "configure", Configure);
+    Nan::SetPrototypeMethod(t, "interrupt", Interrupt);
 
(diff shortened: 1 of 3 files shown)


hrundik:
@springmeyer : my team is also interested in this feature. Do you plan to merge this PR? What is the obstacle of not merging?

panther7:
Hi, please add this to docs API. Thanks

panther7:
Hi, please add this to docs API. Thanks

Issue #525, opened 2015-10-16
SQLite 3.9?


hden:
There a roadmap for supporting 3.9.0 release recently https://www.sqlite.org/releaselog/3_9_0.html?

lvivier:
+1 to 3.9 support

lvivier:
@hden you can close this! 3.9.1 is in, and I have a PR to enable json1 (#538)

PR #532, opened 2015-10-28
Upgrade sqlite 3.8.11.1 -> 3.9.1

lvivier:
ref #525


Mithgol:
Merged, thanks.

Note: in the future we definitely should also start using the `SQLITE_ENABLE_JSON1` compile-time option. That is useful for [managing JSON content](https://www.sqlite.org/json1.html) stored in an SQLite database.

PR #533, opened 2015-10-28
Upgrade mocha

lvivier:
While running tests, I noticed a deprecation message from node:

    (node) child_process: options.customFds option is deprecated. Use options.stdio instead.

Installing the latest mocha gets rid of it.


Mithgol:
Good idea, thanks.

PR #538, opened 2015-11-01
Enable json1 extension by default

lvivier:
ref #532

Enables the [JSON1 extension](https://www.sqlite.org/json1.html) added in 3.9.0, and adds a test to confirm the `json()` function exists.

### DIFF
--- a/deps/sqlite3.gyp
+++ b/deps/sqlite3.gyp
@@ -80,6 +80,7 @@
         'defines': [
           'SQLITE_THREADSAFE=1',
           'SQLITE_ENABLE_FTS3',
+          'SQLITE_ENABLE_JSON1',
           'SQLITE_ENABLE_RTREE'
         ],
       },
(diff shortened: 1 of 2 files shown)


lvivier:
@Mithgol I've only added a simple test to confirm the existence of `json()`, modeled on the fts test. If you think we should do more, just say so.

Mithgol:
In a [Travis CI job](https://travis-ci.org/mapbox/node-sqlite3/builds/88608700) I've seen that the test passes successfully in “our” built SQLite but then fails when the shared (the local system's) SQLite [is tested](https://github.com/mapbox/node-sqlite3/blob/ecb4f3c87c94de881b4b8b7c2e65f636227b7b6b/scripts/build_against_node.sh#L22-L29).

That's normal because that SQLite is not expected to have `SQLITE_ENABLE_JSON1` enabled.

I'm going to [mark the test as pending](http://mochajs.org/#inclusive-tests) for now (fe4812eda4c0a84937451ee8d0b84c312d285bb1).

In the future we might want to skip this test only when the shared SQLite is tested (however, I don't know if Mocha permits this idea).

Mithgol:
I guess I'd better try a change in the system's environment.

Mithgol:
My changes (1fcb843cdfb1abcf5fe03da0f739e0622c043950) seem fine, the test now depends on the SQLite version (“our” or the system's) being tested.

PR #566, opened 2015-12-06
Add install help guide to contributing.md

ErisDS:
- Users are prompted to view this guide before raising an issue. 
- Putting it here as well as in the wiki may help more people to see the info needed to debug their install.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,14 @@
 
 General guidelines for contributing to node-sqlite3
 
+## Install Help
+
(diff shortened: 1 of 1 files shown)


ErisDS:
:confused: not sure why the build is failing here?

Mithgol:
On Travis CI some tests are failing because of `Connection reset by peer` or `couldn't connect to host` while connecting to http://llvm.org/apt/llvm-snapshot.gpg.key

In other words, some problem with the Internet connectivity on Travis. It's not okay, but it happens and we should just ignore it.

On AppVeyor there was one job that couldn't even install Node.

PR #576, opened 2015-12-15
On Mac OS X, never fall back to waiting 1s to aquire transactions

bengotow:
We're building node-sqlite3 against Electron and using it in N1 (http://github.com/nylas/N1). The app opens connections to a sqlite database from several Node processes and we recently discovered that it's waiting intervals of 1000ms to aquire transactions when it encounters `SQLITE_BUSY` because `HAVE_USLEEP` is being erroneously set to false at compile time. (I'm not sure why this is the case—as far as I know, `usleep` has always been available on Mac OS X!)

This PR changes the sqlite gyp config so that `HAVE_USLEEP` is always passed in on Mac OS X. We verified that this corrects our issue in N1 by measuring the wall time it takes to open transactions and observing that it's no longer either `1ms` or `1001ms`.

Given how much compile flags impact sqlite's performance, it might be good to dig in and figure out why this is necessary, and possibly enable it on other platforms via `'cflags': ['-DHAVE_USLEEP=1']`

### DIFF
--- a/deps/common-sqlite.gypi
+++ b/deps/common-sqlite.gypi
@@ -17,6 +17,9 @@
           '-DNDEBUG'
         ],
         'xcode_settings': {
+          'OTHER_CFLAGS': [
+            '-DHAVE_USLEEP=1'
+          ],
           'OTHER_CPLUSPLUSFLAGS!': [
             '-O3',
             '-Os',
@@ -36,6 +39,9 @@
           'NDEBUG'
         ],
         'xcode_settings': {
+          'OTHER_CFLAGS': [
+            '-DHAVE_USLEEP=1'
+          ],
           'OTHER_CPLUSPLUSFLAGS!': [
             '-Os',
             '-O2'


inukshuk:
I am on Arch Linux and node-sqlite3 is not compiled with usleep either (neither is the build picked up by node-pre-gyp). This is a massive performance hit when using multiple connections; any idea why the flag is not picked up? Enabling it manually as suggested by this PR is the only feasible way I can use multiple connections right now.

bengotow:
Hey folks—just wanted to bump this PR. Let me know if there's anything that needs to be done to get this merged.

aguynamedben:
@bengotow This still hasn't been merged (2 years later). I'm building an Electron app that similarly queries SQLite from a few processes. We're seeing a few difficult-to-reproduce crashes. I don't think this will fi the crashes, but is there any reason I *shouldn't* use USLEEP in 2018?

inukshuk:
For what it's worth, we've been compiling SQLite with `HAVE_USLEEP` for an Electron app and haven't had a problem with it over 2 years now. The only real alternative is avoiding concurrent transactions (e.g., have only one thread or process use explicit transactions, and connect all others in read-only mode), because otherwise you will get full second delays all the time, which will frequently escalate to busy timeouts.

aguynamedben:
@inukshuk Thank you very much for this insight. It's very helpful to hear. 🙏

kewde:
Hi,
Will be merging soon, checking whether it passes CI.

inukshuk:
Well, I believe this PR only sets the flag on macOS. But this should really be enabled on Linux too. For our app, I've been adding `'HAVE_USLEEP'` permanently to [sqlite.gyp](https://github.com/mapbox/node-sqlite3/blob/master/deps/sqlite3.gyp#L80) around line 80 and 93.

kewde:
I'm going to enable this for Linux too. It might affect embedded systems a bit, but I doubt they'll be running NodeJS. Any objections?

kewde:
#1090

kewde:
Thank you @bengotow  #1090 has been merged for all platforms.

Issue #591, opened 2016-01-14
npm install sqlite3@3.1.1  err


sea35:
npm ERR! node v4.2.2
npm ERR! npm  v3.5.3
npm ERR! code ELIFECYCLE

npm ERR! sqlite3@3.1.1 install: `node-pre-gyp install --fallback-to-build`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the sqlite3@3.1.1 install script 'node-pre-gyp install --fallback-to-build'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the sqlite3 package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node-pre-gyp install --fallback-to-build
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs sqlite3
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls sqlite3
npm ERR! There is likely additional logging output above.

crapthings:
same here, why

LIXIAOSHUANG:
same here why? 
Same as has a tip that it is "npm ERR! sqlite3@3.1.1 install: `node-pre-gyp install --fallback-to-build  This is most likely a problem with the sqlite3 package,
npm ERR! not with npm itself " 

I had install node-gyp  and node-pre-gyp

ErisDS:
If you're having install issues, please provide all of the information described here: https://github.com/mapbox/node-sqlite3/blob/master/CONTRIBUTING.md#install-help

Without the proper error log and details of your system it's not possible for anyone to see what might be wrong.

crapthings:
osx 10.10.5

brew install version/node012

nvm use 0.12.9
nvm use 5.4.1

12601 silly install node-pre-gyp@0.6.14 /Users/monsterstep/dev/waterlinetest/node_modules/.staging/node-pre-gyp-f4dcdcb45c7320ee317cd6659b86a2c5
12602 info lifecycle node-pre-gyp@0.6.14~install: node-pre-gyp@0.6.14
12603 silly lifecycle node-pre-gyp@0.6.14~install: no script for install, continuing
12604 silly install sqlite3@3.1.1 /Users/monsterstep/dev/waterlinetest/node_modules/.staging/sqlite3-c381cf6804abcab9bef9c649cb254058
[...]

PR #600, opened 2016-02-13
Added a reminder when building on OS X to have the latest Xcode

Darkle:
…Xcode.

re: https://github.com/mapbox/node-sqlite3/issues/591


springmeyer:
Looks like this includes a change to the build system as well as the readme. Therefore it conflicts with #627 as far as I can tell. Please re-open a new pull request that only modifies the README.

PR #608, opened 2016-03-05
npm install can't find node-pre-gyp.

DLLarson:
`node-pre-gyp` was installed properly but there was no path to the actual script.

When building in a cross-tool environment the install choked.

-Dale

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     },
     "scripts": {
         "prepublish":"npm ls",
-        "install": "node-pre-gyp install --fallback-to-build",
+        "install": "node_modules/node-pre-gyp/bin/node-pre-gyp install --fallback-to-build",
         "pretest": "node test/support/createdb.js",
         "test": "mocha -R spec --timeout 480000"
     },


springmeyer:
Npm should put `node-pre-gyp` on the PATH automatically. Tell me more about what your `cross-tool` environment is? Perhaps your NPM installation is broken?

DLLarson:
The build was performed using the Yocto project tools (Fido 1.8.1).

The issue happens when doing `npm install` with modules that require native builds. Normally this all works just fine. We build many packages that include native components. This package busted the builds.

When I investigated I found it was trying to run `node-pre-gyp` but didn't know how to find it.

This was my solution. Yet I noticed in the CI tests it doesn't work on Windows targets.

We've redirected our product build `package.json` to my personal fork of this repo for the near term to get operational again.

-Dale

springmeyer:
> We build many packages that include native components. This package busted the builds.

Npm should be putting node-pre-gyp on PATH just like it does for node-gyp. Perhaps finding node-gyp was busted for you as well and you were working around it since you had a global node-gyp install? A global node-pre-gyp install would also perhaps work around the problem.

DLLarson:
Thanks for the feedback!

I think I need to do some digging yet. 

The Yocto build system provides completely wrapped environments for each task of a build (fetch, patch, configure, compile, install, package...) 

It generally provides its own native tools so as not to rely on the host system's support. So as part of the process a native `node/npm` installation is created and it is used to produce the cross architecture packages. I did actually install `node-pre-gyp` into the native version but the build still didn't find it in the path. `node-sqlite3` is the only package with native components that breaks for us due to `node-pre-gyp`. It could be that none of the other packages use the `pre-gyp` stuff.

-Dale

bedwardly-down:
@DLLarson, I've noticed that this PR is exactly 4 years old to the date and the original repository is no longer available. Any way you could please close it if you are still active on Github? Thanks

daniellockyer:
Closing as it should be longer be relevant 🙂

PR #627, opened 2016-04-18
Added the latest SQLite full text search addon flag (FTS5) to sqlite3…

Darkle:
….gyp


wmertens:
I also do this in #724 

But now that I see that this is dormant since April, I suppose nothing will happen :(

Mithgol:
After #724 is merged this pull request should be closed, or so I think.

PR #630, opened 2016-04-21
Revert "Add support for sqlite3_interrupt"

springmeyer:
Reverts mapbox/node-sqlite3#518

PR #631, opened 2016-04-22
Add support for sqlite3_interrupt (with fixed test)

briangreenery:
This reintroduces #518, fixing the test failure that caused it to be reverted in #630. The test failure was happening because the interrupt test used common table expressions, and common table expressions are not supported on older versions of sqlite. I removed the use of common table expressions in the interrupt test to fix it.


briangreenery:
I'm sorry this caused the build to fail again. I'm investigating this right now.

PR #647, opened 2016-05-11
node-pre-gyp is incorrectly listed as a dependency in the package.json file

cblair:
Should be under the devDependency. Simple enough fix.


springmeyer:
It is required at runtime for installs, so it is a dependency.

cblair:
So this installs stuff at runtime? We're shipping this with a product that has no external internet access, so what will the behavior be?

Issue #41, opened 2011-09-17
update docs to make it clear that exec() always runs in parallel


twobitfool:
I chased my tail for a while because I didn't realize that I should have been using `run()` instead of `exec()`. The later runs in parallel even when using `db.serialize()`.

PR #649, opened 2016-05-15
Fix interrupt test: wait for exec to complete before continuing with test

briangreenery:
@springmeyer I'm sorry to have broken the build twice and caused so much trouble. The reason that the test failed this time is that `db.exec` does not not block other queries from starting in `db.serialize` mode. This is mentioned in issue #41.

I was using `db.exec` to create a test table `t`. Depending on the order that the worker threads did stuff, that might not have happened before the query that used the table `t` was run. I fixed this by waiting for `db.exec` to complete before continuing with the rest of the test.

Sorry again.

### DIFF
--- a/test/interrupt.test.js
+++ b/test/interrupt.test.js
@@ -13,29 +13,34 @@ describe('interrupt', function() {
             for (var i = 0; i < 8; i += 1) {
                 setup += 'insert into t values (' + i + ');';
             }
-            db.exec(setup);
 
-            var query = 'select last.n ' +
-                'from t,t,t,t,t,t,t,t,t,t,t,t,t,t,t,t,t,t,t,t,t,t,t,t,t,t as last';
-
-            db.each(query, function(err) {
+            db.exec(setup, function(err) {
                 if (err) {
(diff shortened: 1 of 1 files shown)


springmeyer:
Thanks for the attentiveness @briangreenery. All green (i'll fix that one appveyor failure later, looks like something odd with node-gyp/node-pre-gyp /cc @BergWerkGIS).

BergWerkGIS:
There's an old version in use that doesn't know about VS2015 yet:

```
versions = _DetectVisualStudioVersions(version_map[version], 'e' in version)
KeyError: '2015'
```


https://ci.appveyor.com/project/Mapbox/node-sqlite3/build/1.0.484/job/23buxwjo7cjt1829#L424

BergWerkGIS:
All green again: https://ci.appveyor.com/project/Mapbox/node-sqlite3/build/1.0.492
Solved it by updating `npm` to `2.15.6` for `node@0.x` builds and leaving others as they are.

springmeyer:
Thank you @BergWerkGIS!

PR #660, opened 2016-06-09
Enable foreign key support by default

babignano:
Enables Foreign Key support by default. 

Foreign Key support is disabled in SQLite 3.x by default and is only enabled via a build flag as in the PR or a pragma setting `PRAGMA foreign_keys = ON;` each time a connection is established - which is not feasible in our app as connections are opened and closed through an ORM (We are using Sails and waterline-sqlite3) so we have no visibility over the connections being opened and closed.

I would suspect this would break a lot of existing apps but at least we can have the conversation.


MartinMuzatko:
Is there any way to help here? I would also like to have foreign keys enabled.

mittalyashu:
Any updates on this PR?

Issue #665, opened 2016-06-18
Node 6.2.2.: TypeError: undefined is not a function


xbolshe:
**Environment:**
Node 6.2.2
SQLite3 3.13.0
node-sqlite3 3.1.4

**Issue:**

> root@quark:/usr/lib# node
> 
> > var sqlite3 = require('sqlite3');
> > undefined
> > var db = new sqlite3.Database(':memory:');
> > undefined
> > undefined:0
> > 
> > TypeError: undefined is not a function
> > root@quark:/usr/lib#

**Code sample:**

```
var sqlite3 = require('sqlite3');
var db = new sqlite3.Database(':memory:');
```

**sqlite3 works well**
root@quark:/usr/lib# sqlite3
SQLite version 3.13.0 2016-05-18 10:57:30
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
sqlite> .open ex1.db
sqlite> create table tbl1(one varchar(10), two smallint);
sqlite> insert into tbl1 values('hello!',10);
sqlite> insert into tbl1 values('goodbye', 20);
sqlite> select \* from tbl1;
hello!|10
goodbye|20
sqlite>

BR,
xbolshe

lvivier:
I couldn't repro this on OS X 10.11.6, with node@6.2.2 and node-sqlite3@3.1.4. Did you do anything special when installing node-sqlite3? What happens if you remove it from node_modules and reinstall?

xbolshe:
I use preinstalled node-sqlite3 to custom Linux image for Intel Galileo board. Previously there was no problem.

I have repeated the same problem with a full recompilations of Linux image with downrgaded packages (Kernel 3.19.8+BusyBox+and so on+NodeJS 6.2.1+SQLite 3.12.1+node-sqlite3 3.1.4). It is strange...

xbolshe:
It seems it is not node-sqlite3 issue. And I close it.

quark login: root
Kernel: Linux quark 3.19.8-yocto-standard #1 Wed Jun 22 22:34:06 MSK 2016 i586 GNU/Linux
Nodejs: v6.2.2, Shared OpenSSL: OpenSSL 1.0.2h  3 May 2016
Thu Jun 23 20:41:22 UTC 2016
Temp: 74°C
root@quark:~# cat > a.js
process.on('uncaughtException', function(error) {
  console.error('Caught error in process!', error)
})

[...]

PR #670, opened 2016-06-26
Add TRY_CATCH_CALL function check and remove Nan::MakeCallback

xbolshe:
Signed-off-by: Eugene Bolshakov pub@relvarsoft.com

This patch fixes 2 issues:
1. There is no check for a callback function presence in TRY_CATCH_CALL
2. Remove Nan::MakeCallback due to https://github.com/nodejs/nan/issues/284 and use Nan::Call

### DIFF
--- a/src/macros.h
+++ b/src/macros.h
@@ -107,7 +107,7 @@ const char* sqlite_authorizer_string(int type);
 
 #define TRY_CATCH_CALL(context, callback, argc, argv)                          \
     if(!callback.IsEmpty() && callback->IsFunction()){                         \
-        (callback)->Call((context), (argc), (argv));                           \
+        Nan::Call((callback), (context), (argc), (argv));                      \
     }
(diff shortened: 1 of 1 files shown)


xbolshe:
It is better to use Nan::Call due to v8 differences.

springmeyer:
@xbolshe - thank you for this fix. It looks like in #665 you also have a testcase that failed before this PR. Can you add that as well?

xbolshe:
If you mean this post https://github.com/mapbox/node-sqlite3/issues/665#issuecomment-228177620, it does not related with node-sqlite3. The issue was related with a MMX support on the target CPU (it does not support MMX). The issue was fixed after a compilation with disabling MMX.

daniellockyer:
This library has since switched away from NAN so this should no longer be relevant - closing for now 🙂

PR #705, opened 2016-09-06
async: remove racey `uv_ref()`/`uv_unref()`

indutny:
`uv_ref()` can't be called from the non-event-loop thread. Furthermore,
it is not necessary in `async.h`, because open `uv_async_t` is already
holding the event loop from exiting.

See: #702

cc @jlongster

_(NOTE: Tests for current master fail on my machine, so please check it with care!)_


indutny:
You are drunk @coveralls :D

tmcw:
👍 beautiful, thank you @indutny !

indutny:
ありがとうございます。

PR #715, opened 2016-09-23
Add basic eslint configuration

tmcw:


tmcw:
Fails because of new eslint not working with old node.

Issue #722, opened 2016-10-12
Upgrade to SQLite to latest release.


rnewman:
`node-sqlite3` ships with sqlite 3.9.1, which was released almost exactly a year ago.

At the very least, this should be the 3.9.3 branch release, which includes two fixes and a backported performance improvement.

Ideally it should be 3.14.2, which includes performance improvements and a year of other work.

springmeyer:
Yes, can upgrade when we make a new release. In the meantime you can easily source compile node-sqlite3 against your own version of libsqlite3 like https://github.com/mapbox/node-sqlite3#source-install

wmertens:
@springmeyer is there anything I can do to help with doing the next release? Issue wrangling, PR reviews, …?

rnewman:
I suggest taking #724 instead.

springmeyer:
Thanks, agree, done in #724

PR #723, opened 2016-10-12
Upgrade sqlite to 3140200

springmeyer:
refs #722


wmertens:
I missed this when I did #724 :) Anyway, 3.15 has row values, see http://use-the-index-luke.com/sql/partial-results/fetch-next-page why that is useful

springmeyer:
replaced by #724, thanks @wmertens

PR #753, opened 2016-12-22
port to be used on z/OS

jBarz:
Use nan 2.5.0
Use xlc options instead of gcc options.
Avoid use of python tar functionality. Use shell instead.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -7,6 +7,11 @@
   "targets": [
     {
       "target_name": "<(module_name)",
+      'conditions': [
+        ['OS!="os390"', {
+          'cflags': [ '-include ../src/gcc-preinclude.h' ],
+        }],
+      ],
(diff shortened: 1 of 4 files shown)


springmeyer:
How can this change be tested on travis? Is there a docker image available for z/OS?

jBarz:
Unfortunately, we don't have community zos systems at the moment but we are working on making some available in the near future

springmeyer:
> Unfortunately, we don't have community zos systems at the moment but we are working on making some available in the near future

Okay, I'm hesitant to accept a PR that I can't confirm works. Is there a way for me to be able to test?

daniellockyer:
This PR is out-of-date with how `node-sqlite3` currently works and builds. If it's still an issue, feel free to open a new PR with the updated method 🙂

PR #755, opened 2016-12-25
Fix OPTIONAL_ARGUMENT_FUNCTION

fanatid:


fanatid:
I think it was done because rest of macros had this.
Right now, I'm not sure what was requirements for, but I believe that wrap any macros argument to parentheses is good practice.

springmeyer:
Thanks. I'll keep close as this will change with N-API: #79

PR #771, opened 2017-02-14
remove 'insert', 'update', 'delete' from supportedEvents

taichi:
there event is needless to configure.

see. https://github.com/mapbox/node-sqlite3/blob/master/src/database.cc#L323


daniellockyer:
Closing as we'll be fixing this in an upcoming release 🙂

PR #799, opened 2017-03-16
Upgrade node-pre-gyp to version 0.6.32

suhaig:
The 0.6.32 version of node-pre-gyp supports npm configuration for CA bundles.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
   },
   "dependencies": {
     "nan": "~2.4.0",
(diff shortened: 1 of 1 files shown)


springmeyer:
Thanks for the contribution. Because `node-pre-gyp` is bundled it will also take republishing a new node-sqlite3 version before this works. We should also upgrade the bundled sqlite3 version at the same time. Would welcome a PR for that and then I'll create a new release.

PR #829, opened 2017-06-07
AppVeyor: use stock node and "ci-scripts"

BergWerkGIS:
* use stock node to build (not deprecated custom Mapbox version of node)
* use `appveyor.yml` and build script downloaded on the fly from https://github.com/mapbox/ci-scripts/tree/master/node
* uses build matrix x86/x64:
  * `node@0.10`
  * `node@4.8.1`
  * `node@6.10.1`
  * `node@8`


daniellockyer:
We're switching away from Appveyor so I'm closing this 🙂

PR #853, opened 2017-07-23
Promise version

sveyret:
I made a promise version of the Database and Statement classes. These versions also have an autoPrefix options to add a prefix to object keys for SQL binding.
Note that the Promise version **does not replace** the classical version, it is composed of 2 new  classes.
I personnaly use these versions (a little, I must say), and they seem to work well.
Please merge it if this can be helpful to you.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,1 +1,5 @@
 module.exports = require('./sqlite3');
+
+var promiseVersion = require('./promise-sqlite3');
+module.exports.PromiseStatement = promiseVersion.PromiseStatement;
+module.exports.PromiseDatabase = promiseVersion.PromiseDatabase;
--- a/lib/promise-sqlite3.js
+++ b/lib/promise-sqlite3.js
@@ -2,6 +2,9 @@
 
 var sqlite3 = require('./sqlite3');
 
+// Prefix for auto-prefix for binding
+var prefix = null;
+
 // Utility functions for Statement
 function promiseFor(fn, args) {
     var object = this;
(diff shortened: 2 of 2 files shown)


sveyret:
Please feel free to comment or ask questions…

wmertens:
So if I understand correctly, this PR adds a JS wrapper that returns promises? I think you could make even wrap it so that calling the regular functions without callbacks returns a Promise…

In any case, do you know https://github.com/kriasoft/node-sqlite ? That one also provides a Promise API. It looks like your version is a bit smaller, so maybe you can create a `sqlite3-promise` package for this?

I must say though that the prefixing feature is a bit arbitrary to be in a library.

sveyret:
@wmertens, thank you for commenting. Your idea of overriding regular functions without callback is good, I'll have a look to see if it is easily doable.

I looked at a some other existing packages before creating this PR, and none were satisfying me. I could write my own (probably I will if you are not interrested by my PR) but I thought, as the Promise pattern is getting more and more popular in Node.js, that it would be good if directly integreted in your package.

Regarding the autoprefix feature, I know it looks totally arbitrary, but I created it for my own needs, because I am using named parameters, prefixed by a colon, which is not easy to use in JS objects. I know one can use the dollar sign, but even then, I find it confusing that `$myvar` refers to the `$myvar` entry and not to the `myvar` one. Using `autoprefix` with a value of ':' allow to use the `:myvar` named variable in the queries and refer it with a `myvar` key in a JS object.

PR #858, opened 2017-08-02
bumped node-pre-gyp version to solve package issue

sandramchung:
Node-pre-gyp versions 0.6.33 and higher fix this issue that causes all of our backend jest test suites to fail silently: https://github.com/mapbox/node-pre-gyp/issues/262


springmeyer:
Solved in #860, and fixed in the v3.1.9 release.

Issue #872, opened 2017-09-11
SQLCipher disappears when electron-rebuild is called


jimcloudman:
I'll preface this by saying that my problem may not be with `node-sqlite3`, but with `electron-rebuild` or `node-pre-gyp` - but for now, this seems to me like the proper place to make a request.

I want to use SQLite in an Electron app with the SQLCipher extension. I've had no problems using SQLite with Electron or SQLCipher with SQLite, but I'm unable to combine the three.

Following [the directions](https://github.com/mapbox/node-sqlite3/blob/master/README.md), I installed SQLite3 from source with the extension attached, like so:

`npm install sqlite3 --build-from-source --sqlite_libname=sqlcipher --sqlite=`brew --prefix` --save--dev`

(Note that this is on MacOS, and that I had installed sqlcipher with Homebrew as well, and had it added to the path. Not sure if that's actually necessary, but I did it anyway.)

Then to make sure that database access worked, I ran a [test script](https://gist.github.com/jimcloudman/84c05df6a3bcbd61cc0b5c9d5edd0ccd) with it. It created an encrypted database - attempting to open it with a tool like DB Browser for SQLite resulted in a password prompt. There's some concurrency issues (the unencrypted test starts before the encrypted test finishes), but it works well enough to confirm SQLCipher's presence. The console output looked like this:

    Encrypted DB access test
    Encrypted DB without password test
    Unable to access without password
    Encrypted access: SUCCESS

However, this doesn't work with Electron. If I try to run my Electron app with just that - without using `electron-rebuild` - I get the following error when I try to run the same code in the app as in my test script:

`Cannot find module '/Users/jim/SourceCode/[APP NAME]/node_modules/sqlite3/lib/binding/electron-v1.7-darwin-x64/node_sqlite3.node`

That makes sense. The module is there, but its binding is for `node-v57-darwin-x64`. Not built for Electron. That's what `electron-rebuild` is for, so I ran the following:

`./node_modules/.bin/electron-rebuild -w sqlite3 -p`

The Electron app was now able to work with sqlite, but could not handle the encrypted database I just created from node:

`SQLITE_NOTADB: file is encrypted or is not a database`

I switched the filename so that it would create a new database. After getting around the concurrency issue, the app's test script showed that I was able to access the new database without a password. From what I could tell, SQLCipher was stripped out of the build - or SQLite3 was reinstalled entirely without the build-from-source & other flags. I'm not sure exactly how `electron-rebuild` works.

[...]

PR #873, opened 2017-09-11
Added note about custom builds & Electron

jimcloudman:
(PR requested per issue #872)

I've added a section to the readme addressing extra steps needed to get sqlite3+sqlcipher running in an Electron app.

I've only got about two days of actual experience working with Electron and Node.js applications, so it may be worth taking a close look at what I've written to address any inaccuracies.


springmeyer:
Thanks, looks good.

PR #883, opened 2017-09-20
added backup function to database object

schraf:
I wanted to have the backup support from sqlite3 exposed in node-sqlite3 for a project I am working on. Here is a link to the documentation for this API: [Using the SQLite Online Backup API](https://sqlite.org/backup.html). I have included a unit test that just makes a backup of an existing database in the test/support directory. I only tested this on my ArchLinux server.

### DIFF
--- a/src/database.cc
+++ b/src/database.cc
@@ -430,7 +430,7 @@ void Database::Work_AfterBackup(uv_work_t* req) {
     Local<Function> cb = Nan::New(baton->callback);
 
     if (baton->status != SQLITE_OK) {
-        EXCEPTION(Nan::New(baton->message.c_str()).ToLocalChecked(), baton->status, exception);
+        EXCEPTION(baton->message, baton->status, exception);
 
         if (!cb.IsEmpty() && cb->IsFunction()) {
             Local<Value> argv[] = { exception };
--- a/src/database.cc
+++ b/src/database.cc
@@ -24,6 +24,7 @@ NAN_MODULE_INIT(Database::Init) {
     Nan::SetPrototypeMethod(t, "parallelize", Parallelize);
     Nan::SetPrototypeMethod(t, "configure", Configure);
     Nan::SetPrototypeMethod(t, "interrupt", Interrupt);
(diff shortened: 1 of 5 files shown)


laysdra7265:
I got an `Error: SQLITE_CANTOPEN: library routine called out of sequence` when trying to use this PR commit to backup sqlite database file.
node version: v6.11.3
system version: centos 7

laysdra7265:
Actually the problem is the path not existed,we need to create the path first.
Or get a `library routine called out of sequence` error without any useful error message.

src/database.cc
```diff
@@ -371,6 +374,83 @@ NAN_METHOD(Database::Interrupt) {
...
+void Database::Work_AfterBackup(uv_work_t* req) {
+    Nan::HandleScope scope;
+
+    BackupBaton* baton = static_cast<BackupBaton*>(req->data);
+    Database* db = baton->db;
+
+    Local<Function> cb = Nan::New(baton->callback);
+
+    if (baton->status != SQLITE_OK) {
+        EXCEPTION(Nan::New(baton->message.c_str()).ToLocalChecked(), baton->status, exception);
```
kewde: > EXCEPTION(baton->message, baton->status, exception);

test/backup.test.js
```diff
@@ -0,0 +1,14 @@
...
+var assert = require('assert');
+
+describe('backup', function() {
+    var db;
+    before(function(done) {
+        db = new sqlite3.Database('test/support/prepare.db', sqlite3.OPEN_READONLY, done);
+    });
+
+    it('backup database', function(done) {
+        db.backup('test/support/backup.db', done);
```
kewde: Seems a bit of a thin test case.
I don't mind improving it if anyone has ideas of what could get wrong.

src/database.cc
kewde: To make it comply with V8 7.1

paulfitz:
@schraf thanks for this, it was helpful. I extended it in #1116 to support the backup api more completely.

Issue #433, opened 2015-03-30
Error: /lib64/libc.so.6: version `GLIBC_2.14' not found


rrs120486:
I added dependency sqlite3 3.0.5 to my node app but it fails with below error.

 Possibly unhandled Error: /lib64/libc.so.6: version `GLIBC_2.14' not found (required by cronus/software/service_nodes/installed-packages/cronus/scripts/node_modules/sqlite3/lib/node_sqlite3.node)

I am using RedHatEnterpriseServer Release: 5.4. Is there any older version of sqlite3 which can run without this error?

Also by any chance is pure js version will be available for this?

bnoordhuis:
I'm seeing the same issue.  It looks like it's just the memcpy symbol:
 
```
$ nm lib/binding/node-v43-linux-x64/node_sqlite3.node | grep GLIBC_2.14
                 U memcpy@@GLIBC_2.14
```

Perhaps a strategically placed .symver directive can help work around that?

```c
#ifdef __GLIBC__
__asm__ (".symver memcpy, memcpy@GLIBC_2.2.5");
#endif
```

springmeyer:
@bnoordhuis - thanks for helping with this! And I see you found my failed previous effort at this (the https://github.com/mapbox/node-sqlite3/blob/master/src/gcc-preinclude.h). I was not able to get it working and that is the background for giving up over at https://github.com/mapbox/node-sqlite3/issues/244#issuecomment-76689979. Its been too long now to recall what did not work previously, but I presume you've confirmed that this fixes things on your system? Out of curiosity what operating system did you hit this on?

bnoordhuis:
> Its been too long now to recall what did not work previously, but I presume you've confirmed that this fixes things on your system? Out of curiosity what operating system did you hit this on?

It works on my x86_64 centos 6 (glibc 2.12) system.  I got a bug report from a customer; they haven't verified it yet but I have no reason to believe it won't work, they run pretty much the exact same configuration.

PR #932, opened 2018-01-09
New: db.all({ sql, rowMode = "array|nest" }), fixes #443

### LINKED ISSUE #443: Joins of tables with duplicate column names returning only last table's columns
Node version: v0.10.28
node-sqlite3 version: 3.0.5

If you have two tables which have, say, ID and name fields:

```
thing: ID, name
```

```
user: ID, name, thingID
```

then join them
```
select thing.id, thing.name, user.id, user.name 
from thing
inner join user on thing.id = user.thingID
```

The returned rowset contains only one pair of ID and name properties which will be the ID and name of the last table, which in this case is user:

```
[
{ id: 1,
  name: "foo" }
]
```

This would be to be expected if the fields in the query were not prefixed with the table name. 
When running the same query in a SQLite DB Browser it returns 
```
id, name, id, name 
```
So it's understandable that if you try to make a JS Object out of that then the later id property would overwrite the first one.

Are we using SQLite wrong? Is it expected for it to strip out table names in returned column names?

We have a workaround, which is to alias each column as tablename_columnname, but this is a performance hit as the result then needs to be processed to remove the tablename_.

dotnil:
This commit changes the Statement callback footprint from `function(err, result)` to `function(err, result, fields)`. The default result structure is changed from `{ field1: value1, field2: value2 }` to `[ value1, value2 ]` to be compliant with the fields array.

The newly added fields is in such format:

```js
[ { name: 'field1', table: 'table1' },
  { name: 'field2', table: 'table1' } ]
```

In this way, when a JOIN query like `SELECT foo.*, bar.* ...` happens, the values in result won't be collided with each other. And the result can be properly processed with the help of fields.

To be backward compliant, both `Statement.prototype.all` and `Database.prototype.all` are overriden. The default `.all(sql, callback)` still works the same, with the result containing objects. The added method type is `.all({ sql, rowMode })` in which the `rowMode` can be `array` or `nest`.

If `rowMode` is array, both `result` and `fields` get returned as is. If `rowMode` is nest, the returned `result` is an array of nested objects in following structure:

```js
[ { table1: { field1: value1, field2: value2 },
    table2: { ... } },
  { table1: { ... },
    table2: { ... } } ]
```

`rowMode = nest` is quite the same as `nestTables = true` mode in the mysql client of Node.js.


dotnil:
Hi maintainers of node-sqlite3, I'm working on a new ORM library for JavaScript called [jorma](https://github.com/dotnil/jorma). I've been trying to add sqlite3 support in the past weekend but with issue #433 found, hence no luck. 

Changing the selects to aliases do solve the JOIN problem but I think it's a bit unnecessary. Hence I've come up with this pr. Currently it's a WIP since the callback of `Statement.prototype.get` hasn't been changed. Besides, I'm not a C++ guru. The changes on `statement.cc` deeply needs review.

dotnil:
@springmeyer This pr is almost finished now. I've changed `.travis-ci.yml` a bit to be update with their latest infrastructure changes.

.travis.yml
```diff
@@ -1,106 +1,22 @@
...
-
-dist: precise
+language: node_js
+node_js:
+ - "node"
+ - "lts/*"
+ - "7"
+ - "6"
+ - "5"
+ - "4"
```
dotnil: I've left out 0.10 and 0.12 on purpose since Node.js is in a much quicker release cycle and the latest LTS major version is already 8.

.travis.yml
```diff
@@ -1,106 +1,22 @@
...
+
+dist: trusty
+os:
+ - linux
+ - osx
```
dotnil: trusty environment has build-essentials installed by default.

deps/sqlite3.gyp
```diff
@@ -78,6 +78,7 @@
         'include_dirs': [ '<(SHARED_INTERMEDIATE_DIR)/sqlite-autoconf-<@(sqlite_version)/' ],
         'defines': [
           'SQLITE_THREADSAFE=1',
+          'SQLITE_ENABLE_COLUMN_METADATA',
```
dotnil: `sqlite3_column_table_name` is behind this flag.

lib/sqlite3.js
```diff
@@ -153,6 +153,77 @@ Database.prototype.removeAllListeners = function(type) {
...
+    };
+
+    var statementMethodOrig = Statement.prototype[name];
+    Statement.prototype[name] = function () {
+        var args = Array.prototype.slice.call(arguments);
+        var callback = args.pop();
+        args.push(callback.length < 3 ? wrapCallback(name, 'object', callback) : callback);
+        return statementMethodOrig.apply(this, args);
+    };
+});
```
dotnil: Override `all`, `each`, and `get` to be backward compatible. If array or nested rows is need, we can now: 

```js
db.all({ sql, rowMode: 'array' }, ...)
db.get({ sql, rowMode: 'nest' }, ...)
```

scripts/build_against_node.sh
```diff
@@ -39,45 +39,3 @@ else
...
-    # broken for some unknown reason against io.js
-    if [[ ${NODE_VERSION:0:4} != 'iojs' ]]; then
-        # test source compile in 32 bit mode against external libsqlite3
-        export NODE_SQLITE3_JSON1=no
-        CC=gcc-4.6 CXX=g++-4.6 npm install --build-from-source --sqlite=/usr  --clang=1
-        npm test
-    fi
-fi
-
-'
```
dotnil: This part is already commented out.

test/named_columns.test.js
```diff
@@ -29,7 +29,7 @@ describe('named columns', function() {
 
     it('should be able to retrieve rowid of last inserted value', function(done) {
         db.get("SELECT last_insert_rowid() as last_id FROM foo", function(err, row) {
-            if (err) throw err;
+            if (err) done(err);
```
dotnil: Haven't look into other test cases, but errors should be passed to `done` callback.

dotnil:
ping @springmeyer

springmeyer:
@dotnil thanks for this PR. I appreciate your attention to the build side as well. I have some outstanding work to do on the build side still  (getting https://github.com/mapbox/node-sqlite3/pull/829 landed and using the mac/linux approach from over at https://github.com/mapbox/node-cpp-skel/), but I'm lacking time right now to do these things.

So, my question for you: does your code changes require changes to the build/travis side? Ideally you could isolate them so that I can finish build changes separately.

As far as review of this code, I'll need help from @kkaefer who is the original author of this module (I'm the release maintainer but am not very familiar with the code).

dotnil:
@springmeyer It's great to hear from you guys. The changes on build scripts are mostly to fix the failing test cases for using clang-3.5 which is too old a version. The commits are separated, one for api changes and the other one for build scripts, so I think I can easily rebase my changes after your work is merged or just discard changes on the build part. I couldn't get them to pass in macOS anyway.

@kkaefer I know the binding changes might be against your will since you've close the issue three years ago (#443). But I do deeply need the changes to happen to make my ORM [jorma](https://github.com/dotnil/jorma) to work without sacrificing code simplicity. Changing the generated sql from `SELECT foo.*` to `SELECT foo.id as "foo:id"` seems trivial but the sql bloats when the tables have to many columns.

dotnil:
`sqlite3_column_table_name()` returns original table name instead of the alias. To fulfill my task, I've added a patch to the bundled SQLite source. Apparently this won't work when compiling against shared SQLite. I've decided to close the pr (for now) and suggest those who need `db.all(sql, { rowMode })` use my fork (temporarily) instead.

rathboma:
Hey @cyjake and @kkaefer, sorry to stir up old memories here, but can I ask why this feature was shelved? This seems immensely helpful and is implemented (pretty much in this exact way) by all the other node database drivers (psql, mysql2, mssql).

I run [Beekeeper Studio](https://github.com/beekeeper-studio/beekeeper-studio) and dupe columns overriding themselves is a particular user frustration.

@kkaefer in particular I'm interested in if you're open to considering the addition of this feature?

Issue #980, opened 2018-05-04
Feature Request - Expose an option to set SQLITE_FCNTL_CHUNK_SIZE


trevershick:
Our C code uses this option to set a 'larger than default' chunk size to help with fragmentation.  However, our Node code is unable to take advantage of this feature.

There is already a `db.configure('busyTimeout', 1000)` feature it would be great if we could `db.configure('chunkSize', 1024*1024)`.

see https://sqlite.org/c3ref/c_fcntl_chunk_size.html
ref (about fragmentation https://stackoverflow.com/questions/13393866/sqlite-wal-performance-improvement)

daniellockyer:
@trevershick Is this still something that is desired? 🙂

trevershick:
> @trevershick Is this still something that is desired? 🙂

I think it's useful but no this is no longer something that's immediately desired.

PR #981, opened 2018-05-04
Add db.configure('chunkSize', <int>);

trevershick:
There was no way (that i could find) to set the SQLITE_FCNTL_CHUNK_SIZE
via the javascript, so i extended the 'configure' method to allow
'chunkSize' to be passed in.

I also added a callback on the configure method. This commit only has
the change for chunkSize but it probably should be extended to all
configure options (the callback that is).

I created a new ConfigureIntBaton that was very specific to the problem of
configuring the db with a numeric value. I did this instead of
overridding the status value in the baton.  It felt cleaner.

Addresses mapbox/node-sqlite3#980


trevershick:
I'm thinking i shouldn't have created this branch from master.  The appveyor builds are borked.

trevershick:
closed due to needing rebase.

PR #986, opened 2018-05-14
Updated npm dependencies

Nostradamos:
Fixes the known vulnerabilities and should also make the david-dm badge green again.
`npm install --build-from-source` works, `npm test` too.


springmeyer:
Thanks, this is now fixed in master and will be part of the v4.0.1 release.

PR #998, opened 2018-06-07
Sqlite 3.24.0

bruce-one:
Update SQLite version to 3.24.0.

(And add a simple test verifying that upsert is present, which is a bit pointless... But, eh....)


bruce-one:
(I think the test failure is unrelated to the change? :-s )

bruce-one:
Thanks @springmeyer! :-)

PR #1002, opened 2018-06-20
Upgrade Travis to use the Trusty environment

palfrey:
Travis' Precise environment is now deprecated, and so we need to upgrade to the newer option.


springmeyer:
The benefit of using precise is that the GLIBC version of binaries created on precise will be compatible across more linux versions, including older Centos and RHEL. So, I plan to keep using precise until travis fully removes support.

Issue #1061, opened 2018-10-22
Update coverage


mapsam:
Coverage commands are failing with the following error:

```
./scripts/build_against_node.sh: line 20: cpp-coveralls: command not found
```

We should:

1. ✅ ~~[short term] rip this coverage out to avoid breaking tests~~ completed in #1062 
1. [long term] update to use codecov

cc @springmeyer

springmeyer:
> [short term] rip this coverage out to avoid breaking tests

👍 

long term I've hoped to find time to forward port node-cpp-skel to node-sqlite, which would be a medium-size effort but would get codecov for free.

mapsam:
PR for short-term solution over at #1062

daniellockyer:
Closing this for now - but I will re-add coverage 🙂

PR #1062, opened 2018-10-23
remove coverage

mapsam:
This removes coverage, which was breaking the builds. It's a short-term solution so others don't get unnecessarily stuck. Long-term solution is described in #1061 

cc @springmeyer @kewde


mapsam:
Thanks @springmeyer - going to merge, but no need for a release here!

Issue #1065, opened 2018-10-25
Is there a flag to *statically* link SQLCipher?


aguynamedben:
Hi, I'm building an Electron app with SQLCipher. Following the README instructions for SQLCipher, everything worked in development, but when I deployed the new version of my app, it failed because SQLCipher was dynamically linked and my users haven't installed SQLCipher on their computer.

Resources:
* [Original issue that led to SQLCipher support](https://github.com/mapbox/node-sqlite3/issues/195)
* [node-sqlite3's binding.gyp](https://github.com/mapbox/node-sqlite3/blob/master/binding.gyp)
* [How to tweak binding.gyp to statically link instead of dynamically link](https://github.com/nodejs/node-gyp/issues/328#issuecomment-24335081)
* [Maybe some other useful info on static vs. dynamic linking via gcc flags](https://stackoverflow.com/a/938947/3516664)

springmeyer:
Yes, it is reasonable to add an option for this. The link to the static library will need to be completed to be absolute to work right.

aguynamedben:
@springmeyer Thanks for the feedback.

I'm still testing this within the context of my app, but I submitted a PR that allows `--sqlite-static-library=/full/path/to/libsqlcipher.a` to be defined. Let me know if something looks off in there or you prefer another way of passing the static library path.

aguynamedben:
@springmeyer My final effort toward making static linking for Electron+SQLCipher users is here: https://github.com/mapbox/node-sqlite3/pull/1069 It's a docs-only PR that explains the problem.

Providing docs to explain this issue to Electron+SQLCipher users seems like a better solution than to further complicate logic in binding.gyp. Plus I'm a n00b when it comes to gyp, compiler flags, and linkers.

This is an issue isolated to Electron+SQLCipher users, and it's simple to overwrite ./node_modules/sqlite3/binding.gyp with a modified version, rebuild, and package.

PR #1066, opened 2018-10-26
Static linking option for rebuilding (`--sqlite-static-library`)

aguynamedben:
Allow users to pass an absolute path to a SQLite static library by
providing a `--sqlite-static-library` flag. This will cause the
binding to be compiled with static linking instead of dynamic
linking. The purpose of allowing this is to allow non-standard
versions of SQLite (i.e. SQLCipher) to be distributed to machines
without the SQLCipher dynamic libraries (i.e. an Electron app that
uses SQLCipher).

Discussion: https://github.com/mapbox/node-sqlite3/issues/1065

Example of static linking (Homebrew-installed SQLite on macOS)
```
npm rebuild sqlite3 --build-from-source --sqlite-static-library=`brew --prefix`/lib/libsqlcipher.a --sqlite=`brew --prefix`
```

Review notes:
* I'm new to node-gyp and gyp, so please let me know if there's a better way to do things within binding.gyp.
* I haven't tested this on Windows or Linux.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ In the case of MacOS with Homebrew, the command should look like the following:
 
 #### Electron static linking
 
-By default, the node-sqlite3 bindings compiles with dynamic linking to SQLite. If you're building an
+By default, the node-sqlite3 bindings compile with dynamic linking to SQLite. If you're building an
 Electron app, chances are when you package and distribute your app the user will not have the
 SQLCipher dynamic library on their computer and your app will fail to launch when sqlite3 is
 imported.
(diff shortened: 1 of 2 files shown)


aguynamedben:
After 1.5 days of messing with various issues around these gyp changes, I realized I could just modify LDFLAGS to point to libsqlcipher.a and that will cause static linking. It took me a while because I'm a n00b at messing with gyp, compiler flags, and linking. I think I understand it now though.

```
export LDFLAGS="-L`brew --prefix`/opt/sqlcipher/lib/libsqlcipher.a"
```

I will close this PR and submit a doc-only PR that explains this to other SQLCipher users that need to statically link SQLCipher for distribution with an Electron app.

aguynamedben:
Superseded by https://github.com/mapbox/node-sqlite3/pull/1069

PR #1067, opened 2018-10-27
Update SQLCipher section of README with static linking instructions for Electron

aguynamedben:
Related to:
* https://github.com/mapbox/node-sqlite3/issues/1065
* https://github.com/mapbox/node-sqlite3/pull/1066

Simple instructions to modify LDFLAGS seems more graceful than hacking up binding.gyp to support a non-core use case via a flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,18 +122,18 @@ Remember the following:
 
 Visit the “[Using Node modules](https://github.com/rogerwang/node-webkit/wiki/Using-Node-modules)” article in the node-webkit's wiki for more details.
 
-## Building for sqlcipher
+## Building for SQLCipher
 
-For instructions for building sqlcipher see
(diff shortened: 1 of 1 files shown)


aguynamedben:
Gah, this isn't quite right, it requires more overrides in binding.gyp. Will submit another docs-only PR with instructions on how to replace binding.gyp with a static linked version.

aguynamedben:
(I keep forgetting to uninstall SQLCipher as the final stage of testing, as this is an issue that only happens at run-time when SQLCipher isn't installed, but I need to have SQLCipher installed to build the bindings.)

aguynamedben:
Superseded by https://github.com/mapbox/node-sqlite3/pull/1069

PR #1078, opened 2018-11-10
fix #1032: add constants for file open

### LINKED ISSUE #1032: Creating shared memory database fails
Hello.

I'm trying to create a database in memory, run some queries on it to create tables, and then attach that same database to another one.

According to the sqlite3 documentaiton, you can create shared memory databases.

https://www.sqlite.org/inmemorydb.html see the "In-memory Databases And Shared Cache" section.

I tried something like that, but it fails :

```$ node
> var sqlite3 = require('sqlite3').verbose();
undefined
> var db = new sqlite3.Database('file::memory:?cache=shared');
undefined
> events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: SQLITE_CANTOPEN: unable to open database file

```

I could always attach the memory database first and then run queries on it, but since I'm reusing queries I already ran elsewhere, I'd have to copy/paste them to change table names (`table1` to `memorydb.table1`) for example.

This works fine when using a sqlite3 client in bash.

Would it be possible to implement it ?

gms1:
This PR only propagates some missing constants for file open to JavaScript:

please see #1032 for a description of possible use cases

[Flags For File Open Operations](https://www.sqlite.org/c3ref/c_open_autoproxy.html)


kewde:
Looks good, thanks!
Seems like a CI failure, I'll restart it.

kewde:
I've added a sample test for the shared database, to test a small aspect of it. Ideally we'd have more testing of the shared database but it seems like a more niche feature. On the other hand, I haven't seen tests specifically aimed at read-only databases either.

kewde:
@gms1 could you shine some light as to why the database creation is failing in the test I added? That seems to be the default way to open databases of different modes without the URI protocol.

gms1:
@kewde 
many thanks! 
something like that should work:
```
'file:sqlite3orm?mode=memory&cache=shared'
```
quoting from https://www.sqlite.org/inmemorydb.html
In-memory databases are allowed to use shared cache if they are opened using a URI filename. If the unadorned ":memory:" name is used to specify the in-memory database, then that database always has a private cache and is this only visible to the database connection that originally opened it.

gms1:
please review the added/modified tests

gms1:
hm all tests are passing during the first run at least, but only on linux, the "shared memory database" tests are failing during the second run: "SQLITE_ERROR: no such access mode: memory"
whats the difference between these two?

kewde:
The test pass for the SQLite3 package that has been bundled with the application.

```

kewde:
@gms1 
I think we can get the version of SQLite3 using `select sqlite_version();` and then only run the test if the version supports it.

gms1:
@kewde 
please review

kewde:
Thank you for the contribution @gms1!

Issue #1084, opened 2018-11-20
Ambiguous python executable name


FFY00:
https://github.com/mapbox/node-sqlite3/blob/master/deps/sqlite3.gyp#L58

Here we are calling the `python` executable, this can point to either python2 or python3 depending on the distribution. Should be changed to `python2`.

kewde:
Closing in favor of #1089

PR #1089, opened 2018-12-04
Allow specifying the python to use.

bruce-one:
Alternative fix for #1084.

Just if of interest :-)


kewde:
Thanks again @bruce-one !

PR #1090, opened 2018-12-04
HAVE_USLEEP=1 for all platforms

kewde:


kewde:
Enabling it for all builds doesn't seem to affect Windows as its value gets overwritten.
Windows will remain using the sleeping intervals, although still faster.

https://www.autoitscript.com/forum/topic/170514-is-sqlitedll-compiled-with-have_usleep1/

PR #1129, opened 2019-02-24
builds sqlite3 with version 3270100

prejrdev:
With this version you can create an in memory db and use `VACUUM INTO 'somefile.db';` to export the database to the fs without using the backup api. Also prior versions do not support this use case.


kewde:
Thanks, doing the update to 3.28 along with the new release for node 12.

PR #1144, opened 2019-04-15
Allow custom magic to be specified through --sqlite_magic

danielgindi:


kewde:
Thank you @danielgindi.

PR #1149, opened 2019-04-18
[WIP][stalled] ARM Build in Docker in Travis

### LINKED ISSUE #418: Automated ARM builds
I would be happy to build arm6 and arm7 bins for both sqlite3 and sqlcipher.

v0.10 and v1.2 work

v0.11 and <= v1.1 fail

v0.12 I haven't tested.

I have internet-accessible PIs connected to a few routers that I can offer for such a task.

tkurki:
This PR is at this stage a proof of concept of building ARM binary in Docker with Travis.

The first commit comments out the normal build.

The second commit has the real beef:
- use the default dist (Docker seems to be not available in `precise`)
- an additional dummy build target for ARM
- script steps to install ARM support and the docker command to build & publish inside the prebuilt arm32v7/node container off Docker hub

This works in Travis up to ["Package appears valid"](https://travis-ci.org/tkurki/node-sqlite3/builds/521698951#L2529) but naturally publish stage fails without the S3 credentials.

Is this something that you would consider merging? 

I have not tested the resulting binary yet.

If finalised and merged fixes #418.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: false
 
 language: generic
 
-dist: precise
+# dist: precise
 
 git:
   depth: 10
@@ -15,8 +15,14 @@ addons:
   #   packages:
   #    - clang-3.5
 
+services:
+  - docker
+
 matrix:
   include:
+    # ARM with Docker
+     - os: linux
+       env: NODE_VERSION="8" USE_DOCKER=1
     #  # Linux
     #  - os: linux
     #    compiler: clang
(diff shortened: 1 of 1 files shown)


rzr:
will travis have enough resource to crossbuild ?

tkurki:
Please see the linked Travis log file. Needs an extra `travis_wait` but seems to build it.

kewde:
Thanks,

The PoC is certainly interesting but it does need to follow the same flow as the other release procedures.
That means it needs to run the tests before even publishing a binary.

tkurki:
Sure, but I stopped here before spending time on something that might not be merged because there’s something fundamentally off with it.

Any other input as to how to proceed? Make tests run, spread to the supported node versions - anything else I should be aware of?

kewde:
I'm going to have to consult the Travis docs a bit further, adding the ARM architecture will require additional builds for the different NodeJS versions. Electron also has an ARM version, so if we also want to provide those binaries we are looking at doubling the load of the CI essentially. 

It does not seem to me that there is a max limit on the amont of jobs, rather just a max limit for each job.

kewde:
To make it more production friendly:
* Restore the  `.travis.yaml` file
* Move ARM specific setup code into a new shell script unders the /scripts
* Have the original file `build_against_node.sh` run within that docker/qemu (may need modification for the naming of the releases etc)

tkurki:
I have not abandoned this, just low on the todo list for the time being, so if anybody else feels like giving it a shot go right ahead.

tkurki:
See https://github.com/mapbox/node-sqlite3/pull/1362

PR #1152, opened 2019-04-27
tests: await for backup to complete before finish

kewde:
The CI is returning segfaults, I believe because the database objects are finished before the backup fully completes. I've added awaits.
@paulfitz

### DIFF
--- a/test/backup.test.js
+++ b/test/backup.test.js
@@ -158,7 +158,7 @@ describe('backup', function() {
                 if (err) throw err;
                 var backup = db.backup('test/tmp/backup.db', 'temp', 'main', true, function(err) {
                     if (err) throw err;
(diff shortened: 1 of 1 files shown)


kewde:
I'm sure it's a timing issue, closed this PR because it contains keywords (async, await) which are not available in all versions of JS.

paulfitz:
Ah, I'm sorry to see this code causing trouble.  For those two specific tests, I'd seen an issue, but thought it was actually a sqlite version thing (hence the `sqlite3.VERSION_NUMBER < 3007011 ? it.skip : it` checks).  I'll take another shot at trying to replicate it locally.  Worst case, it would be reasonable to turn off the functionality this is testing and these two tests, 99% of use-cases won't need it.

PR #1154, opened 2019-04-27
tests: wait for the backup to complete before finishing it

kewde:


kewde:
After quite a bit of experimenting (thanks @tecnovert) we've come to the conclusion that the latest (currently 3.26) built-in version does not experience any of the segfault issues. 

To the best of our knowledge, there is a timing issue within `assertRowsMatchDb` when called with the same db.

kewde:
AppVeyor is experiencing some unrelated networking error. Merging.
```
gyp ERR! stack Error: This is most likely not a problem with node-gyp or the package itself and
gyp ERR! stack is related to network connectivity. In most cases you are behind a proxy or have bad 
gyp ERR! stack network settings.
```

PR #1186, opened 2019-07-10
electron-builder: attempt more lightweight builds

kewde:


daniellockyer:
I don't think this is needed any more 🙂

Issue #1148, opened 2019-04-18
the 'request' dependency is creating a black hole on my hard disk


coolaj86:
But more seriously, I'm a bit concerned that there are now about 100 dependencies to node-sqlite3 whereas before there were only... 2? 3?

Would you be open to a PR that replaces `request` with a less risky option?

I've written an API-compatible drop-in replacement for `request` that has 0 dependencies.

I don't want credit or to increase my download count or even to promote it, but it is something that is far more secure (by virtue of not having 100 dependencies as attack vectors) and you can literally drop it in as a replacement for 99% of use cases:

[@root/request.js](https://git.rootprojects.org/root/request.js/src/branch/master/index.js)

It's so small it could just be copy-pasted. No need to even add it as a dependency (again, not trying to promote myself or increase downloads, I just want node-sqlite3 to be "safe" again).

With the various occasions this past year of popular projects that have senseless numbers of dependencies being comprised and having malware installed on people's computers, I'm just very wary of projects with kitchen-sink dependencies like request (no offense to Mikeal or his excellent work).

ryanburnette:
I was able to resolve an image of the `node_modules` directory @solderjs is referring to.

![node_modules](https://user-images.githubusercontent.com/2252601/56334607-fe266b80-6166-11e9-9843-0f17d0a36a8a.jpg)

RomanBurunkov:
I even haven't manage to find for what purpose 'request' was included into dependencies.

kewde:
Thanks for the issue, I fully agree with the reasoning.
I did not like this either. I did a bit of investigating into this issue as well a while ago.
cfr #1132

This was a fix that was before my time but I'm open for a PR, but it can't break the issue it was aimed to resolve.

Blitzsturm:
I dug into this a while back and can see request isn't actually used for anything; it appears to be included by mistake... As a correction I now use specifically version 4.0.1.  It's relatively minor but I don't like importing an instantly bloated package if I'm deploying to a lean server.

kewde:
> I dug into this a while back and can see request isn't actually used for anything; it appears to be included by mistake... As a correction I now use specifically version 4.0.1. It's relatively minor but I don't like importing an instantly bloated package if I'm deploying to a lean server.
> 
> This is a pretty easy fix... I don't know how it's been there for 10+ months.

It is not included by accident. It was done on purpose, in fact node-pre-gyp will use it.
[...]

PR #1192, opened 2019-07-24
Get rid of "request"

grabus:
Request is not used anywhere, but it creates a security risk and brings tons of its own dependencies.
Also, its contributors don't want to release new version for a year.
We definitely should get rid of this.
issue #1148


bedwardly-down:
Thus has already been addressed in PR https://github.com/mapbox/node-sqlite3/pull/1287

PR #1200, opened 2019-08-03
User functions

### LINKED ISSUE #140: User Defined Functions?

rlindner81:
Based on PR https://github.com/mapbox/node-sqlite3/pull/448 by @wbyoung
This also fixes #140 

* I updated the syntax to work with the current `nan`
* I also added a cleanup mechanism very loosely inspired by https://github.com/python/cpython/blob/master/Modules/_sqlite/connection.c#L848

Without the cleanup, the `async_t` handlers used here will always be considered active by libuv making node run forever. I suspect that is also what @springmeyer observed on the travis builds.

NOTE: I have very little experience programming in node C/C++ addons, so a thorough review would be appreciated.


kewde:
Thank you for your contribution @rlindner81, I will take a look at this.

rlindner81:
@kewde If you need any input to expedite the process, I can do some more work on this.

SamuelBrucksch:
@kewde, is there any progress on this? I would really appreciate this functionality.

RobinMalfait:
Is there anything I can help with to get this functionality merged?

rlindner81:
@RobinMalfait I don't think this will be merged. I never heard back from @kewde

RobinMalfait:
Thanks for your work though @rlindner81 !

RobinMalfait:
@rlindner81 I am currently your branch for our tests, however in our case Jest keeps on running without never cleanly exiting. 

image

I wonder if you might know why this happens and eventually how to fix this. I am no C++ programmer, but if you happen to know _where_ to start looking (the --detectOpenHandles also never finishes 🙃 ) that would be highly appreciated!

Thanks in advance!

rlindner81:
Hi @RobinMalfait, these types of bugs are hard to track down in my experience. Start by double-checking that the problem is caused by the changes in this branch. Then I would track it down further by commenting large amounts of code and seeing if that fixes it... You have to do a kind of binary search for the culprit that way. 
The code in this branch is unfortunately hopelessly outdated and I will not invest more work in it sorry. I close the PR to make that clearer.

SamuelBrucksch:
Too bad, thanks for your work Richard. Guess we have to look for something else then.

PR #1233, opened 2019-10-27
Use util.promisify.custom to improve promisify support of db.run()

### LINKED ISSUE #658: Problem when promisify `run`
Currently the callback argument for `run` only accept one parameter, namely `err`, and `lastID` and `changes` can be get from `this` inside the callback. When promisify using tools like bluebird, things like `lastID` are lost. It would be nice to have those passed to the callback as the second argument, and this is also a nodejs convention I believe.

kimamula:
Fixes #658

Currently, promisifying `db.run()` results in a loss of information of `lastID` and `changes`.
In this PR, I customized the promisification of `db.run()` using `util.promisify.custom` to solve the problem.

### DIFF
--- a/lib/sqlite3.js
+++ b/lib/sqlite3.js
@@ -79,19 +79,21 @@ Database.prototype.run = normalizeMethod(function(statement, params) {
     return this;
 });
 if (typeof util.promisify === 'function') {
-    Database.prototype.run[util.promisify.custom] = function() {
-        var runThis = this;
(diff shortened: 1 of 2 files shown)


kimamula:
It is likely the tests fails due to https://github.com/nodejs/node-gyp/issues/1933

daniellockyer:
@kimamula Would you mind rebasing on top of `master` so CI can be run? 🙂

PR #1237, opened 2019-10-28
prebuilt: add electron 6.1 & 7

kewde:


kewde:
I will have to bump the version of Node for the build process of electron 7.x because it's using the `async`keyword which is available as part of Node.js since version 7.10.1. It's currently using 6.

jackple:
Is there a certain time about  bumping the version of Node?

kewde:
@jackple I'm working on it today. I have two bugs to resolve before I can do a release.

jackple:
@kewde Thanks a lot! I'm using electron 6.x and I've got a few problems that only upgrading to 7.x could help me.

PR #1255, opened 2019-11-21
Use minimum supported node version for Electron 7

EmilioCristalli:
Not sure if this change is right, but wondering if this is the reason the CI was failing with Electron 7.
Looks like the minimum supported node version for Electron 7 is Node 8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -132,7 +132,7 @@ matrix:
      # electron Linux
      - os: linux
        compiler: clang
-       env: NODE_VERSION="6" ELECTRON_VERSION="7.0.0"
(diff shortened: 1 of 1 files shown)


EmilioCristalli:
@kewde any idea why some Travis jobs failed when I ran them from my repo but worked fine from this PR? (using the same commit)

For example:
https://travis-ci.org/mapbox/node-sqlite3/jobs/615116333
https://travis-ci.org/EmilioCristalli/node-sqlite3/jobs/615105993

It's like the one from my repo was missing xvfb, but can't figure out why, since it should be using the same Travis config

kewde:
Thank you for the contribution.

The appveyor errors are unrelated, will merge after the osx checks are done.

PR #1256, opened 2019-11-22
prebuilt: add missing x86 versions for electron 4.1 & 4.2

kewde:


daniellockyer:
Closing as I don't think this is relevant any more 🙂

Issue #1249, opened 2019-11-14
builds: deprecating support for version that have reached end of life


mourner:
Currently Travis builds for this repo take an excessively large amount of time:

> Ran for 59 min 31 sec
> Total time 4 hrs 38 min 17 sec

This is due to it being tested on 22 (!) Electron versions and 19 (!) Node versions. This seems quite redundant — many of these reached End of Life (e.g. Node <8), or are intermediary development versions that were succeeded by stable ones (e.g. all odd Node versions except the last one).

I think this needs a thorough cleanup so that it doesn't fill up Travis resources so much, leaving other projects hanging in the queue for a long time. Even for a project of this scale, I would expect 3-4 times less jobs.

cc @kewde @mapsam

EmilioCristalli:
maybe, as a first step, just reduce the node versions to those listed as supported in the Readme? 
(`The sqlite3 module works with Node.js v4.x, v6.x, v8.x, v10.x, v11.x and v12.x.`)

Will open a PR just in case you guys think it's worth it

kewde:
The CI environment is also the main source for the prebuilt binaries of the releases.
I can not do a release for a version without having it in the CI.

So removing them, means not supporting them at all. It's hard to decide which versions to stop supporting without knowing which version isn't being used anymore.

Perhaps the mapbox team could tell us the statistics, how many downloads a particular version gets from the S3 bucket. That way we can make a more informed decision.

EmilioCristalli:
that's a good point.

yeah, I was thinking about not supporting them at all. Looking at the Readme it seemed like they weren't supported anyway. And I think Node doesn't maintain them anymore

But, agreed, would be nice to see some stats

mourner:
@kewde personally I think that regardless of download stats, `node-sqlite3` should only support Node versions that are officially maintained, and drop pre-building binaries for unmaintained/legacy versions (shouldn't be a big deal). But it would be nice to get a second perspective — @springmeyer what do you think?

springmeyer:
I agree @mourner - and the way to ensure that work without breaking downstream users unnecessarily (who may be using older node or electron versions) is to do a major semver bump when older versions are removed.

sgillies:
@springmeyer @mourner if we can't reduce the size of the matrix now, what would you think about reducing the number of concurrent builds? Other projects can get backed up behind this one. Would 2-3 be an okay cap?

kewde:
[...]

Issue #1228, opened 2019-10-09
Update Travis dist from Precise to Xenial


siikasmaa:
The Linux distribution (Precise) defined in `.travis.yml` is being decomissioned from Travis CI. The following alert can be found when triggering a job:

>This job is running on our Precise environment, which is in the process of being decommissioned. Please update to a newer Ubuntu version by specifying dist: xenial in your .travis.yml.

More about this:
https://docs.travis-ci.com/user/reference/overview/#deprecated-virtualization-environments

kewde:
For the time being, we will remain on trusty as much as possible as it produces builds with a lower version requirement for GLIBC.

daniellockyer:
This should be no longer relevant 🙂

PR #1257, opened 2019-11-22
Remove unsupported node versions

EmilioCristalli:
Trying to make CI builds faster, I removed the old node versions that are not listed in the Readme, assuming they are not really supported anymore.

Related to #1249 

Feel free to close this PR if you guys think it's still worth running these versions in the CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,13 +50,6 @@ matrix:
             packages: [ 'clang-3.5', 'libstdc++-4.9-dev']
      - os: linux
        compiler: clang
-       env: NODE_VERSION="9"
-       addons:
-         apt:
-            sources: [ 'ubuntu-toolchain-r-test','llvm-toolchain-precise-3.5', 'gcc-multilib', 'g++-multilib', 'libsqlite3-dev:i386' ]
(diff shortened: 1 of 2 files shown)


bedwardly-down:
According to https://github.com/mapbox/node-sqlite3/issues/1228#issuecomment-546904869, @kewde is wanting to keep Travis setup for as long as possible on an older LTS version of Ubuntu, so unless otherwise stated, let's safely assume that older nodejs releases may still be needed until upgrading is absolutely necessary.

Off the top of my head, I can think of a small handful of reasons for this: 

1. There are production level servers still on Trusty due to stability reasons which means that Node-Sqlite3 needs to still be compatible with them. 
2. Upgrading to new releases of packages on servers must be handled carefully due to new bugs being introduced, database formats being deprecated, etc. Basically a lot of things that could lead to catastrophic events for customers and hosts alike. 
3. If current versions of dependencies aren't a massive security hazard, why upgrade if you don't have the resources available to handle any issues that arise from it?

kewde:
closed by cc4b598c8cbaaa128c0528fa4a2075bfca17a5f6

PR #1263, opened 2019-12-19
refactor: use bindings to load .node file

codebytere:
This PR alters `.node` file loading to use `bindings`. I'd like to make this change because it enables Webpack to understand how to handle this module, a need which would apply to, for example, Electron apps which both need to use native node modules and use Webpack.

This should not disrupt any existing functionality, and `bindings` itself only has one dependency, which makes this change have minimal impact as far as added dependencies are concerned.

Thank you!

### DIFF
--- a/lib/sqlite3.js
+++ b/lib/sqlite3.js
@@ -1,7 +1,5 @@
-var binary = require('node-pre-gyp');
 var path = require('path');
-var binding_path = binary.find(path.resolve(path.join(__dirname,'../package.json')));
-var binding = require(binding_path);
+var binding = require('bindings')('node_sqlite3.node');
 var sqlite3 = module.exports = exports = binding;
 var EventEmitter = require('events').EventEmitter;
 
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
   "dependencies": {
     "nan": "^2.12.1",
     "node-pre-gyp": "^0.11.0",
+    "bindings": "^1.5.0",
     "request": "^2.87.0"
   },
   "devDependencies": {


kewde:
Thank you for your contribution @codebytere 

I'm closing and re-opening the PR to trigger the CI again because I noticed some false positives.
The failures in the CI seem unrelated to any of  these changes.

codebytere:
@kewde i think they're flakes too - please let me know if there's anything else i need to take care of for this PR otherwise :)

kewde:
I've had the time to inspect this a little bit deeper. 

The behavior of `bindings` is to  check all the possible locations that a native addon would be built at, and returns the first one that loads successfully. 

The behavior of `find() in node-pre-gyp` builds up the path using information from package.json, the node runtime, etc. In short, find() will guarantee that it tries to load the correct version.

If everything goes well and *only* the right bindings are downloaded/built then this shouldn't cause any problems. But I've seen many situations where people (myself included) have updated their package (e.g from v4.1.0 to v4.1.1) but the bindings (.node files) of the older version are not cleaned up and still available in the directory. They are not cleaned up properly. I suspect this PR may cause the loading of older bindings.

codebytere:
@kewde do you see a path forward here in that event?

kewde:
@codebytere I just hope the bindings library fails to load .node bindings that were compiled for an older ABI version. I will test this out, if it fails to load the older bindings than the behavior remains unchanged.

I don't see it as a blocker for this pull request though.

kewde:
@codebytere
Does using #1268 also solve the webpack issue?
I'm more leaning to accepting that one simply because:
1. no additional dependency
2. same behavior as before

Please let me know if I'm missing something.
Thank you.

PR #1265, opened 2019-12-19
updated node-pre-gyp version to ^0.14.0

### LINKED ISSUE #1264: Installing with --target=<target-node-version> results with "Unsupported target version" for targets >= 11
When trying to install sqlite3 with `--target` arg, with targets >= 11.x.x, node-pre-gyp throws an error. It seems that there's an issue with node-pre-gyp v0.11.0. When forcing it to work with node-pre-gyp 0.14.0 (latest at the moment) - it works and selects the closest available target. 
It's important to mention that in node-pre-gyp 0.11.0, the --fallback-to-build doesn't work for this case - so if fails the installation altogether.

To reconstruct, simply run `npm install sqlite3 --target=12.13.1`

mceIdo:
Fixes #1264 - Installing with --target=<target-node-version> results with "Unsupported target version" for targets >= 11

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
   },
   "dependencies": {
     "nan": "^2.12.1",
-    "node-pre-gyp": "^0.11.0",
+    "node-pre-gyp": "^0.14.0",
     "request": "^2.87.0"
   },
   "devDependencies": {


bedwardly-down:
@mceIdo, I see one core issue with upgrading to this release of this package: Node 4 and 5 are still supported for Clang-3.5 in the travis config file. The minimum version that this upgrade supports is 6, so that would need to be one of the first things that needs to be resolved.

mrsimb:
please review this pull request
i need latest node-pre-gyp to install prebuilt binaries from local path due to corporate firewall issues.

bedwardly-down:
The problem is: kewde, the maintainer, tried upgrading to that version only to have it break something. He never documented what happened but did close the PR he tried setting up. I don’t know enough about how all of this works yet to be able to know what the issue was but do know it’s currently not viable yet.

Could you test this out, @mrsimb? I’ll get around it to myself once able. Just working extra to make sure the Covid-19 pandemic doesn’t make me homeless if my city goes under lockdown.

mrsimb:
The real problem atm is that every pull request is failing because some of appveyor ci jobs always fail.
These jobs try to download binaries for electron but the server it downloads from sends 403 error.
I think someone with access to ci config should just disable these jobs so we could continue to receive updates of this library. People who depend on binaries build by these jobs will have to build them from source.

bedwardly-down:
> The real problem atm is that every pull request is failing because some of appveyor ci jobs always fail.
> These jobs try to download binaries for electron but the server it downloads from sends 403 error.
> I think someone with access to ci config should just disable these jobs so we could continue to receive updates of this library.

There might be other issues with it too outside of the tests failing. I’m not a full maintainer but am here to assist with getting things working properly. I just don’t have full permissions yet and the main maintainers are trying to get things going as quickly as they can within their means. Thanks for your patience and we’ll get this sorted out as quickly as we’re able.

marcpo:
Would it be possible to get this change reviewed/fixed soon ?

MasterOdin:
It should be noted that node-pre-gyp fixed a double run error so this upgrade would be awesome as otherwise it makes installing this project normally (`npm install`) impossible for me if a remote source is not available.

Ideally, I'd love to see this backported to the 4.2.0 release series as well.

PR #1313, opened 2020-04-24
fix: [BUG] yarn 2 pnp issues #1312

### LINKED ISSUE #1312: [BUG] yarn 2 pnp issues
The library requires the node-gyp package when there is no existing pre built binaries but it is not declared as a dependency. 

Because yarn 2 uses PlugNPlay, it is not allowed to use non declared dependencies, causing my CI tools to fail

Temp fix is to add to .yarnrc.yml
```yaml
packageExtensions:
  sqlite3@*:
    dependencies:
      node-gyp: "*"
```

regevbr:
fix #1312

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,8 @@
   },
   "dependencies": {
     "nan": "^2.12.1",
-    "node-pre-gyp": "^0.11.0"
+    "node-pre-gyp": "^0.11.0",
+    "node-gyp": "^6.1.0"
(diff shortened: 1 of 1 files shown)


kewde:
Triggered CI - I'll take your word for it that this fixes the issue.
Expected to be included 5.0.0

regevbr:
Thanks @kewde !

PR #1317, opened 2020-04-27
Don't require throw to add trace info for verbose mode

### LINKED ISSUE #1316: Verbose doesn't work with pormises
If I try to use Promises with `node-sqlite3`, `verbose()` doesn't work.

consider the following example:

```js
const sqlite3 = require('sqlite3').verbose();

function db_run(sql){
  return new Promise((resolve, reject) => {
    db.run(sql, err => {
      if (err) {
        console.log('callback has error ...');
        reject(err);
      } else {
        resolve()
      }
    });
  });
}

db_run('update non_existent_table set id=1').catch(err => {
  console.log('promise rejected ...');
  throw err;
});
```

This works as expected and I get the promise rejected, but the error I get doesn't have any trace information :

```
$ node app
callback has error ...
promise rejected ...
(node:15600) UnhandledPromiseRejectionWarning: Error: SQLITE_ERROR: no such table: non_existent_table
(node:15600) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:15600) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

The docs says that I should `throw` the error, if I throw it in `Promise.catch()` I don't get the trace,
But if I throw it in `db_run` in the callback instead of `reject(err)`, I get the trace information but the promise never rejects because `throw err` happens in an async callback so the promise doesn't catch it, neither does placing a `try/catch` around the callback.

```
$ node app
callback has error ...
C:\sqlite-test\node_modules\sqlite3\lib\trace.js:36
                    throw err;
                    ^

Error: SQLITE_ERROR: no such table: non_existent_table
--> in Database#run('update non_existent_table set id=1', [Function])
    at C:\sqlite-test\app.js:6:8
    at new Promise (<anonymous>)
    at db_run (C:\sqlite-test\app.js:5:10)
    at Object.<anonymous> (C:\sqlite-test\app.js:18:1)
←[90m    at Module._compile (internal/modules/cjs/loader.js:955:30)←[39m
←[90m    at Object.Module._extensions..js (internal/modules/cjs/loader.js:991:10)←[39m
←[90m    at Module.load (internal/modules/cjs/loader.js:811:32)←[39m
←[90m    at Function.Module._load (internal/modules/cjs/loader.js:723:14)←[39m
←[90m    at Function.Module.runMain (internal/modules/cjs/loader.js:1043:10)←[39m {
  errno: ←[33m1←[39m,
  code: ←[32m'SQLITE_ERROR'←[39m,
  __augmented: ←[33mtrue←[39m
```

`p.s.` working on a PR.

mmghv:
Fixes #1316 

Currently, `verbose()` adds trace info to the error object **only** if you `throw` the error by catching it inside a try/catch :

https://github.com/mapbox/node-sqlite3/blob/afa22076ba13ef3bc6afac892d9ed46a7ec36fc2/lib/trace.js#L18-L28

This doesn't work with promises as discussed in the linked issue because throwing in an async callback doesn't get caught by the promise or a wrapping try/catch.

this PR solves this issue by simply adding the trace info outside try/catch by checking the error object which is always the first argument.

### DIFF
--- a/lib/trace.js
+++ b/lib/trace.js
@@ -15,17 +15,14 @@ function extendTrace(object, property, pos) {
         var cb = arguments[pos];
         if (typeof arguments[pos] === 'function') {
             arguments[pos] = function replacement() {
-                try {
-                    return cb.apply(this, arguments);
-                } catch (err) {
-                    if (err && err.stack && !err.__augmented) {
-                        err.stack = filter(err).join('\n');
-                        err.stack += '\n--> in ' + name;
-                        err.stack += '\n' + filter(error).slice(1).join('\n');
-                        err.__augmented = true;
-                    }
-                    throw err;
+                var err = arguments[0];
+                if (err && err.stack && !err.__augmented) {
+                    err.stack = filter(err).join('\n');
+                    err.stack += '\n--> in ' + name;
+                    err.stack += '\n' + filter(error).slice(1).join('\n');
+                    err.__augmented = true;
                 }
+                return cb.apply(this, arguments);
             };
         }
         return old.apply(this, arguments);
--- a/test/verbose.test.js
+++ b/test/verbose.test.js
@@ -1,0 +1,60 @@
+var sqlite3 = require('..');
+var assert = require('assert');
+
+var invalid_sql = 'update non_existent_table set id=1';
+
+var originalMethods = {
+    Database: {},
+    Statement: {},
+};
+
+function backupOriginalMethods() {
+    for (var obj in originalMethods) {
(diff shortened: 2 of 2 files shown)


kewde:
Thank you for the contribution @mmghv.
I am currently working on release 5.0.0 and will include this tonight.

PR #1347, opened 2020-06-23
Fix bad comparison of c string

mtgto:
(While investigation of #1335) I found some bad c-string comparison.
C++ compiler already warns this bug in CI.
example https://travis-ci.org/github/mapbox/node-sqlite3/jobs/685282346#L683

IMO, this bug doesn't really seem to cause negative consequences, thanks to awesome modern C++ compiler optimization.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
   },
   "peerDependenciesMeta": {
     "node-gyp": {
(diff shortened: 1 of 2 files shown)


kewde:
Thanks again @mtgto

PR #1361, opened 2020-07-21
Update node-gyp in package.json

ono-max:
Hey guys!

I update the version of `node-gyp` from `3.x.x` to `7.0.0` because the version of node-gyp is old in node_modules of node-sqlite3 when I use it.

If you have a request, please let me know. Thank you!


package.json
```diff
@@ -50,15 +50,15 @@
     "mocha": "^5.2.0"
   },
   "peerDependencies": {
-    "node-gyp": "3.x"
+    "node-gyp": "7.0.0"
```
cschanaj, line 53: Is there any reason to hard-code version `7.0.0` instead of `7.x`  ?

mirceanis:
This closes #1415

PR #1367, opened 2020-08-03
Fix worker threads crash

### LINKED ISSUE #1365: Crash when used with worker threads
The following code will crash:

```javascript
const { isMainThread, Worker } = require('worker_threads');
const sqlite3 = require('sqlite3');
if (isMainThread) new Worker(__filename);
function main() {
  db.serialize(() => {
    db.run('CREATE TABLE lorem (info TEXT)');
    while (true) {
      const stmt = db.prepare('INSERT INTO lorem VALUES (?)');
      stmt.finalize();
    }
  });
  db.close();
}
main();
```

Trace:

```
FATAL ERROR: Error::ThrowAsJavaScriptException napi_throw
 1: 0x100a59d5d node::Abort() (.cold.1) [/opt/local/bin/node]
 2: 0x100085ca7 node::FatalError(char const*, char const*) [/opt/local/bin/node]
 3: 0x100085e10 node::OnFatalError(char const*, char const*) [/opt/local/bin/node]
 4: 0x100085cb0 node::OnFatalError(char const*, char const*) [/opt/local/bin/node]
 5: 0x100061f67 napi_open_callback_scope [/opt/local/bin/node]
 6: 0x10457118a Napi::Error::Fatal(char const*, char const*) [/Users/mohamed/Downloads/node-sqlite3/lib/binding/napi-v3-darwin-x64/node_sqlite3.node]
 7: 0x10457457d Napi::Error::ThrowAsJavaScriptException() const [/Users/mohamed/Downloads/node-sqlite3/lib/binding/napi-v3-darwin-x64/node_sqlite3.node]
 8: 0x104588247 node_sqlite3::Statement::Statement(Napi::CallbackInfo const&) [/Users/mohamed/Downloads/node-sqlite3/lib/binding/napi-v3-darwin-x64/node_sqlite3.node]
 9: 0x104592b86 Napi::ObjectWrap<node_sqlite3::Statement>::ConstructorCallbackWrapper(napi_env__*, napi_callback_info__*)::'lambda'()::operator()() const [/Users/mohamed/Downloads/node-sqlite3/lib/binding/napi-v3-darwin-x64/node_sqlite3.node]
10: 0x104592900 Napi::ObjectWrap<node_sqlite3::Statement>::ConstructorCallbackWrapper(napi_env__*, napi_callback_info__*) [/Users/mohamed/Downloads/node-sqlite3/lib/binding/napi-v3-darwin-x64/node_sqlite3.node]
11: 0x10004f7f1 v8impl::(anonymous namespace)::FunctionCallbackWrapper::Invoke(v8::FunctionCallbackInfo<v8::Value> const&) [/opt/local/bin/node]
12: 0x1001dca35 v8::internal::FunctionCallbackArguments::Call(v8::internal::CallHandlerInfo) [/opt/local/bin/node]
13: 0x1001dbe15 v8::internal::MaybeHandle<v8::internal::Object> v8::internal::(anonymous namespace)::HandleApiCallHelper<true>(v8::internal::Isolate*, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::FunctionTemplateInfo>, v8::internal::Handle<v8::internal::Object>, v8::internal::BuiltinArguments) [/opt/local/bin/node]
14: 0x1001db8dd v8::internal::Builtin_Impl_HandleApiCall(v8::internal::BuiltinArguments, v8::internal::Isolate*) [/opt/local/bin/node]
15: 0x100781fd9 Builtins_CEntry_Return1_DontSaveFPRegs_ArgvOnStack_BuiltinExit [/opt/local/bin/node]
16: 0x100716921 Builtins_JSBuiltinsConstructStub [/opt/local/bin/node]
Abort trap: 6
```

The culprit seems to be the `HasInstance` method of `Database` which is called in `Statement::Statement`. The constructor that it's checking against isn't context-specific. The issue is fixed when the `HasInstance` check is removed. I'm not sure how it can be replaced with a context-aware version.

@jschlight

mohd-akram:
Fixes #1365.

### DIFF
--- a/src/database.h
+++ b/src/database.h
@@ -21,7 +21,9 @@ class Database;
 
 class Database : public Napi::ObjectWrap<Database> {
 public:
+#if NAPI_VERSION < 6
     static Napi::FunctionReference constructor;
+#endif
     static Napi::Object Init(Napi::Env env, Napi::Object exports);
 
     static inline bool HasInstance(Napi::Value val) {
@@ -29,7 +31,13 @@ public:
         Napi::HandleScope scope(env);
         if (!val.IsObject()) return false;
         Napi::Object obj = val.As<Napi::Object>();
+#if NAPI_VERSION < 6
         return obj.InstanceOf(constructor.Value());
+#else
+        Napi::FunctionReference* constructor =
+            env.GetInstanceData<Napi::FunctionReference>();
+        return obj.InstanceOf(constructor->Value());
+#endif
     }
 
     struct Baton {
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ local.env
 .eslintrc.js
 setup.sh
 /build-tmp-napi-v3
+/build-tmp-napi-v6
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,7 @@ matrix:
      # electron Linux
      - os: linux
        compiler: clang
-       env: NODE_VERSION="12" ELECTRON_VERSION="8.2.0"
+       env: NODE_VERSION="12.13.0" ELECTRON_VERSION="8.2.0"
        dist: trusty
        addons:
          apt:
@@ -83,7 +83,7 @@ matrix:
             packages: [ 'clang-3.5', 'libstdc++-4.9-dev']
      - os: linux
        compiler: clang
-       env: NODE_VERSION="12" ELECTRON_VERSION="8.1.0"
+       env: NODE_VERSION="12.13.0" ELECTRON_VERSION="8.1.0"
        dist: trusty
        addons:
          apt:
@@ -91,7 +91,7 @@ matrix:
             packages: [ 'clang-3.5', 'libstdc++-4.9-dev']
      - os: linux
        compiler: clang
-       env: NODE_VERSION="12" ELECTRON_VERSION="8.0.0"
+       env: NODE_VERSION="12.13.0" ELECTRON_VERSION="8.0.0"
        dist: trusty
        addons:
          apt:
@@ -99,7 +99,7 @@ matrix:
             packages: [ 'clang-3.5', 'libstdc++-4.9-dev']
      - os: linux
        compiler: clang
-       env: NODE_VERSION="12" ELECTRON_VERSION="7.2.0"
+       env: NODE_VERSION="12.8.1" ELECTRON_VERSION="7.2.0"
        dist: trusty
        addons:
          apt:
@@ -107,7 +107,7 @@ matrix:
             packages: [ 'clang-3.5', 'libstdc++-4.9-dev']
      - os: linux
        compiler: clang
-       env: NODE_VERSION="12" ELECTRON_VERSION="7.1.0"
+       env: NODE_VERSION="12.8.1" ELECTRON_VERSION="7.1.0"
        dist: trusty
        addons:
          apt:
@@ -115,7 +115,7 @@ matrix:
             packages: [ 'clang-3.5', 'libstdc++-4.9-dev']
      - os: linux
        compiler: clang
-       env: NODE_VERSION="12" ELECTRON_VERSION="7.0.0"
+       env: NODE_VERSION="12.8.1" ELECTRON_VERSION="7.0.0"
        dist: trusty
        addons:
          apt:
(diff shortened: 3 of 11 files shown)


jschlight:
This property of `binary` in the `package.json` file creates a native addon that can be used only in those versions of Node that support N-API v6 or later:

```
"napi_versions": [
  6
]
```

I suggest changing this value to:

```
"napi_versions": [
  3,
  6
]
```

This will cause `node-pre-gyp` to fire off two builds: One for N-API v3 and another for N-API v6. For each build, the C preprocessor symbol `NAPI_VERSION` will be set to `3` and `6` respectivly. The `NAPI_VERSION` symbol can then be used for conditional compilation. The original code can then be compiled when `NAPI_VERSION` is `3`, and the new code compiled when `NAPI_VERSION` is `6`.

This will permit older versions of Node to still use the existing functinality without the context-awareness. 

There are more details in the [`node-pre-gyp` documentation](https://github.com/mapbox/node-pre-gyp#n-api-considerations).

mohd-akram:
@jschlight Thanks, updated. Just a guess, but It seems that the `get_napi_version` check [here](https://github.com/mapbox/node-pre-gyp/blob/v0.12.0/lib/util/napi.js#L31) yields an incorrect value when using Electron headers, hence the failures.

jschlight:
Yes. `node-pre-gyp` relies on the version number of the Node process running the build to determine the feasible N-API versions to build. The Electron builds are being performed using Node.js v12.18.3 which supports N-API v6. Therefore, `node-pre-gyp` is requesting a build for N-API v6 even though the build is using Node.js headers for previous Node.js versions that do not support N-API v6. 

`node-pre-gyp` supports a `--target` option that lets you specify a Node.js version. The documentation is [here](https://github.com/mapbox/node-pre-gyp#options). I'm not sure if the N-API build code in `node-pre-gyp` supports this option. But looking at the comments in the code for `get_napi_version` does not make me optimistic. 

As a first step, you might consider adding a `--target` option to the Electron builds that specifies the Node.js version of the headers the binary is being built with. 

If the builds still fail, I can look into this deeper when I return to the office the week of August 17. It may require an update to `node-pre-gyp` which I'm happy to pursue.

mohd-akram:
I changed the Node.js versions in the CI to match Electron's versions.

Helper:
```
curl -Ls https://electronjs.org/headers/index.json |
jq 'sort_by(.version) | .[] | select(.version | inside("8.2.0 8.1.0 8.0.0 7.2.0 7.1.0 7.0.0 6.1.0 6.0.0")) | {version,node}'
```

src/database.h
```diff
@@ -21,15 +21,23 @@ class Database;
...
         Napi::Env env = val.Env();
         Napi::HandleScope scope(env);
         if (!val.IsObject()) return false;
         Napi::Object obj = val.As<Napi::Object>();
+#if NAPI_VERSION < 6
         return obj.InstanceOf(constructor.Value());
+#else
+        Napi::FunctionReference* constructor =
+            env.GetInstanceData<Napi::FunctionReference>();
+        return obj.InstanceOf(constructor->Value());
```
davedoesdev, line 39: You should check `constructor` isn't null before dereferencing it.

mohd-akram: It should never be null though. It's set in the `Init`.

davedoesdev: What if `HasInstance` is called for a `val` which doesn't have instance data in its `env`?

mohd-akram: Ah, the val's env would be the same env that we add the instance data to. The environment is shared by all the objects.

davedoesdev: OK, there's no way of passing an object with a different env - that would have to be from a different worker thread which would require serialization.

kewde:
Thank you.
5.0.2 :+1:

seeiuu:
I'm using threads.js with sqlite3 (through express) inside a worker thread and it's working fine for me so far. But if I call, in the same express project, through a simple route, express crash.

cirosantilli:
I and others have observed related crashes at: https://github.com/mapbox/node-sqlite3/issues/1381 BTW unfortunately.

PR #1407, opened 2020-12-06
Support calling close() on cached.Database()

rhansen:
Introduce a refcount for cached `Database` entries so that calling `close()` on the `Database` returned from `cached.Database()` does not invalidate existing or future uses of the same cached `Database`.
    
This is a backwards-incompatible change because it changes the number of times `close()` must be called before the underlying `Database` is closed.

### DIFF
--- a/lib/sqlite3.js
+++ b/lib/sqlite3.js
@@ -34,20 +34,32 @@ sqlite3.cached = {
 
         var db;
         file = path.resolve(file);
-        function cb() { callback.call(db, null); }
 
-        if (!sqlite3.cached.objects[file]) {
-            db = sqlite3.cached.objects[file] = new Database(file, a, b);
+        let cacheEntry = sqlite3.cached.objects[file];
+        if (!cacheEntry) {
+            cacheEntry = sqlite3.cached.objects[file] = {refCount: 0};
+            db = cacheEntry.db = new (class extends Database {
+                close(cb) {
+                    if (--cacheEntry.refCount <= 0) {
+                        delete sqlite3.cached.objects[file];
+                        super.close(cb);
+                    } else if (typeof cb === 'function') {
+                        process.nextTick(() => { cb.call(this, null); });
+                    }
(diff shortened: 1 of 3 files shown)


Imran-imtiaz48 (commented):
1. **Caching Mechanism:** The implementation uses a caching mechanism for SQLite database connections, which can reduce the overhead of repeatedly opening and closing database files.
2. **Reference Counting:** Reference counting with `refCount` helps efficiently manage resources by ensuring that the database connection is only closed when it's no longer referenced. This is useful for preventing resource leaks and optimizing memory use.
3. **Callback Management:** The conditional callback check ensures that the provided callback is executed appropriately, either on `open` or using `nextTick` for asynchronous consistency.

**Suggestions for Improvement:**
1. **Error Handling:** It would be beneficial to add error handling in places where the callback is invoked to handle unexpected cases gracefully. This would be especially useful in `close` when `super.close` is called.
2. **Code Readability:** To improve readability, consider renaming variables like `a` and `b` to more descriptive names. For instance, `a` could be `flags`, and `b` could be `callback`, making it easier to understand the code's purpose.
3. **Class Extraction:** The `Database` class extension within the `if (!cacheEntry)` block could be extracted outside of the conditional structure, enhancing readability and potentially making it easier to test.
4. **Singleton Pattern Compliance:** Since this code uses a singleton-like pattern for database instances, a further improvement would be to abstract the logic into a `getOrCreateDatabaseInstance` method. This would encapsulate the instance creation and retrieval logic more explicitly.

**Minor Tweaks:**
- **Commenting:** A few more inline comments explaining the purpose of each conditional block would improve code maintainability.
- **Edge Case Testing:** Make sure to test scenarios where `refCount` fluctuates significantly or when the cache entry is deleted during high-concurrency database usage.

Overall, this is a well-thought-out caching and resource management approach for database connections. A few adjustments would enhance the maintainability and robustness of the implementation.

PR #1439, opened 2021-02-07
Switch from mocha to jest.

lygstate:
Jest comes along with code coverage, that's good

Signed-off-by: Yonggang Luo <luoyonggang@gmail.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "@mapbox/cloudfriend": "^1.9.0",
     "aws-sdk": "2.x",
     "eslint": "3.5.0",
-    "mocha": "^5.2.0"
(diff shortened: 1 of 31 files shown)


soryy708:
Mocha can be used with [nyc](https://www.npmjs.com/package/nyc) which is a CLI of [istanbul](https://istanbul.js.org/), that is used behind the scenes of Jest.
Also, CI says the build fails.

daniellockyer:
Hey @lygstate 👋🏻 I think we're ok with Mocha for now. Code coverage is one of the things on my list to add but that's still doable with Mocha.

I'm going to close this for now but I may re-open if we change our mind 🙂

Issue #1438, opened 2021-02-02
How to compile sqlite3 with SQLITE_ENABLE_MATH_FUNCTIONS flag


trevorhreed:
I'm hoping to install sqlite3 via npm but with the math functions (cos, sin) enabled. According to the [sqlite3 website](https://www.sqlite.org/draft/lang_mathfunc.html), these math functions are only available if sqlite3 is compiled with the DSQLITE_ENABLE_MATH_FUNCTIONS flag. Despite reading [this](https://www.npmjs.com/package/sqlite3#user-content-custom-builds-and-electron) section of the node-sqlite3 instructions, I can't figure out how to pass this flag in to the build process.

kewde:
Hi,

I expect this to land after 30th of march when a new version of SQLite3 is released upstream.

Jeremy-Doneghue:
Any updates on this release?

paridhi-agr:
Please release the node-gyp dependency fix. It is causing severe security vulnerabilities

kollinmurphy:
Is anyone actively working on this? It'd be a very nice feature to have.

paulfitz:
I think this happened in https://github.com/mapbox/node-sqlite3/pull/1507

daniellockyer:
This is released in `sqlite3` v5.0.3 🙂

PR #1447, opened 2021-03-03
amalgamate extension-functions.c (sqrt,cos,stdev,etc)

kaizhu256:
this pr is intended to improve node-sqlite3's datascience-capabilities out-of-the-box, e.g.:

```js
let db;
db = require("sqlite3");
db = new db.Database(":memory:");
db.serialize(function () {
    db.exec(`
CREATE TABLE data1 (val REAL);
INSERT INTO data1 VALUES (1.0), (2.0), (3.0), (4.0), (5.0), (6.0), (7.0);
    `);
    db.all("SELECT STDEV(val) AS std FROM data1;", function (ignore, rows) {
        console.log(rows[0].std);
        // output - 2.160246899469287
    });
});
```

admit the pr is a little hacky, and am open to suggestions to make it more clean ^^;;;

### DIFF
--- a/deps/extension-functions.c
+++ b/deps/extension-functions.c
@@ -1,0 +1,1947 @@
+/*
+This library will provide common mathematical and string functions in
+SQL queries using the operating system libraries or provided
+definitions.  It includes the following functions:
+
+Math: acos, asin, atan, atn2, atan2, acosh, asinh, atanh, difference,
+degrees, radians, cos, sin, tan, cot, cosh, sinh, tanh, coth, exp,
+log, log10, power, sign, sqrt, square, ceil, floor, pi.
+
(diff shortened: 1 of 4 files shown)


kewde:
First and foremost, thank you for taking the time to contribute, it's very much appreciated.

I was taking a look at #1438 which seems to reference that this extension is already present in the draft source code of sqlite3?
SQLite3 3.35 will include this functions (30th March) it will be released. 

@kaizhu256 are you interested in PR'ing an alternative solution (upgrade to 3.35 + set flags + linker flags).

That would be a pull request that I would be more willing to merge, given that any time spent reviewing this PR will be undone by the next release.

kaizhu256:
oh wasn't aware.  yea its cleaner.  will look into it when i have time

kewde:
Hi, closing this pull request in favor of #1507

PR #1471, opened 2021-06-14
Improve select performance

mscdex:
I made a few other changes along the way while I was digging in the code, but the main change here is fetching the column names only once instead of for every field for every row which not only measurably improves performance but should use less memory as well.

With the included benchmark of selecting 100k rows, `db.each()` shows ~35% improvement on my machine and `db.all()` shows ~24% improvement. Better measurements could probably be made using the `benchmark` module though.

Running `/usr/bin/time -v node benchmark/select.js` shows a reduction in max RSS of ~5%.

This was all tested on Linux with node v10.22.1.


daniellockyer:
@mscdex Very cool! Would you be able to rebase the PR on top of `master` so CI can run?

Would you also mind splitting apart the code improvements and the performance improvement (w/ benchmark script) into separate PRs? It's harder to review the PR when multiple things are going on 🙂

mscdex:
RE: code splitting, I'm not sure what you're asking for exactly as there are already several commits to this PR which separate the improvements from the performance improvement.

Unfortunately I'm not really invested in this PR anymore as I ended up just writing my own sqlite addon instead. If someone else wants to pick this up in a new PR, I say go for it.

PR #1501, opened 2021-09-23
Full support 64 bit integers with BigInt

renatoalencar:
Well this is not full ready yet, but I'd like to bring some discussions:

1. To use SQLite int64 instead of int32, because JS Number type is always 64 bits;
2. That would make possible to serialize BigInt in queries;
3. Also would make possible to use numbers (with BigInt) between 2^53 and -2^53.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -49,7 +49,7 @@
         "src/node_sqlite3.cc",
         "src/statement.cc"
       ],
-      "defines": [ "NAPI_DISABLE_CPP_EXCEPTIONS=1" ]
+      "defines": [ "NAPI_VERSION=<(napi_build_version)", "NAPI_DISABLE_CPP_EXCEPTIONS=1" ]
     },
     {
       "target_name": "action_after_build",
--- a/binding.gyp
+++ b/binding.gyp
@@ -49,7 +49,7 @@
         "src/node_sqlite3.cc",
         "src/statement.cc"
       ],
-      "defines": [ "NAPI_VERSION=<(napi_build_version)", "NAPI_DISABLE_CPP_EXCEPTIONS=1" ]
+      "defines": [ "NAPI_DISABLE_CPP_EXCEPTIONS=1" ]
     },
     {
       "target_name": "action_after_build",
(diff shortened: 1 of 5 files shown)


juanrgm:
This seems fine, why is it locked?

daniellockyer:
@renatoalencar Would you be able to rebase the PR? 🙂

renatoalencar:
> @renatoalencar Would you be able to rebase the PR? 🙂

Sure! I'll try to do that later today.

daniellockyer:
@renatoalencar Thanks! I've had a quick look and I think the PR might need some gating on `NAPI_VERSION`, because I'd like to maintain N-API v3 support for Node 11/13 (preferably we can avoid doing a breaking release right now). But it LGTM otherwise 🙂

renatoalencar:
> @renatoalencar Thanks! I've had a quick look and I think the PR might need some gating on `NAPI_VERSION`, because I'd like to maintain N-API v3 support for Node 11/13 (preferably we can avoid doing a breaking release right now). But it LGTM otherwise 🙂

I was expecting something like that, as I've mentioned my first implementation wasn't really ready. But I added the extra checks and seems to be working.

Thank you

juanrgm:
In the meantime, I forked your repository at https://github.com/juanrgm/node-sqlite3/tree/int64-support and forced the build from source (https://github.com/juanrgm/node-sqlite3/commit/881d6bfa5809b8b6dd3e7e75d7f23a8ba70341a8) for installing with npm easily:

```sh
npm install https://github.com/juanrgm/node-sqlite3/tarball/int64-support
```

```js
const sqlite3 = require('sqlite3').verbose();

db.serialize(() => {
    db.run("CREATE TABLE t (v INTEGER)");
    const v = 8876343627091516928n
    const stmt = db.prepare("INSERT INTO t VALUES (?)");
    stmt.run(v)
    stmt.finalize();
    db.each("SELECT v FROM t", (err, row) => {
        console.log({ row });
        if (row.v !== v)
            throw new Error('Not equal')
    });
});

db.close();
```

PR #1503, opened 2021-09-29
Update package.json

blubaustin:
Fixes a circular dependency error, and a security vulnerability having to do with tar.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+This is a fork of node-sqlite3 due to them not updating it. This fork fixes a circular dependency error and security vunerability with tar. I will try to keep this fork up to date the best I can.
+
 Asynchronous, non-blocking [SQLite3](https://sqlite.org/) bindings for [Node.js](http://nodejs.org/).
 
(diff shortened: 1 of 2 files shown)


samuelms1:
How does this fix the tar vulnerability?  I'm not saying it doesn't, but it isn't obvious to me.  I can see that node-sqlite3@5.0.2 => node-gyp@3.8.0 => tar@v2.2.2

I see this MR updates `node-addon-api` which I guess updates node-gyp?

README.md
```diff
@@ -1,3 +1,5 @@
+This is a fork of node-sqlite3 due to them not updating it. This fork fixes a circular dependency error and security vunerability with tar. I will try to keep this fork up to date the best I can.
```
samuelms1, line 1: Does it make sense to keep this comment ^^?  This isn't a fork.

blubaustin:
Because the older node version is what was causing the issue. You can try compiling it yourself, and then the audit command actually works. When before it didn't work. It wanted to downgrade the sqlite version for some reason.

PR #1510, opened 2021-10-11
fix method declarations to conform with napi default for methods

alexanderfloh:
Fix native method implementation to use the proper default options for methods (see https://github.com/nodejs/node-addon-api/issues/811 for details).

This has the beneficial effect that it allows instrumenting the sqlite3 API for tools like Open Telemetry or other APM tools.

Note: instrumentation was possible before sqlite3 v5 and the switch to NAPI.


alexanderfloh:
@kewde can you please consider this for merging? Thanks!

daniellockyer:
@alexanderfloh Would you be able to rebase the PR? 🙂

PR #1511, opened 2021-10-13
Fix propagation of async hook ids through callbacks

alexanderfloh:
Use `Call` instead of `MakeCallback` in order to retain the correct async-id through callbacks.

This is also recommended in the Node.js documentation:
> Note it is not necessary to use `napi_make_callback` from within a `napi_async_complete_callback`; in that situation the callback's async context has already been set up, so a direct call to `napi_call_function` is sufficient and appropriate. 
(https://nodejs.org/api/n-api.html#n_api_napi_make_callback)


alexanderfloh:
@kewde can you please consider this for merging? Thanks!

daniellockyer:
@alexanderfloh Would you be able to rebase the PR? 🙂

PR #1528, opened 2021-11-22
Enable extended SQLite error codes

bpasero:


daniellockyer:
@bpasero Would you mind rebasing the PR so CI can be run? 🙂

bpasero:
@daniellockyer done!

PR #1539, opened 2022-01-20
Default to python3 for gyp

MasterOdin:
If `npm_config_python` is not set, then the build process for sqlite3 defaults to searching for `python` command. Using this command is ambiguous at best on what version of python will run, and at worst does not exist (e.g. on alpine). It is encouraged that scripts use an explicit python version (`python2` or `python3`) to ensure best compatibility with supported python platform. In this case, `python3` works perfectly, and so no reason to use the EOL `python2`.

### DIFF
--- a/deps/sqlite3.gyp
+++ b/deps/sqlite3.gyp
@@ -60,7 +60,7 @@
           'outputs': [
             '<(SHARED_INTERMEDIATE_DIR)/sqlite-autoconf-<@(sqlite_version)/sqlite3.c'
           ],
-          'action': ['<!(node -p "process.env.npm_config_python || \\"python\\"")','./extract.py','./sqlite-autoconf-<@(sqlite_version).tar.gz','<(SHARED_INTERMEDIATE_DIR)']
+          'action': ['<!(node -p "process.env.npm_config_python || \\"python3\\"")','./extract.py','./sqlite-autoconf-<@(sqlite_version).tar.gz','<(SHARED_INTERMEDIATE_DIR)']
         }
       ],
       'direct_dependent_settings': {


MasterOdin:
Hm, I guess this breaks the Windows test build at least as it only ever provides `python.exe` in the environment. However, defaulting to `python` breaks various Linux distros, and may at some point break on macOS:

> WARNING: Python 2.7 is not recommended.
> This version is included in macOS for compatibility with legacy software.
> Future versions of macOS will not include Python 2.7.
> Instead, it is recommended that you transition to using 'python3' from within Terminal.

Not sure what the ideal solution here is to get this to work consistently out of the box for everyone, without encoding some sort of "python look-up" into node.

MasterOdin:
Seems like sqlite3 may not be installable on a fresh mac running 12.3 as `python` might be removed? https://developer.apple.com/documentation/macos-release-notes/macos-12_3-release-notes#Python

rwaldron:
Can confirm @MasterOdin's prediction. 


```
gyp info spawn args [ 'BUILDTYPE=Release', '-C', 'build' ]
  ACTION deps_sqlite3_gyp_action_before_build_target_unpack_sqlite_dep Release/obj/gen/sqlite-autoconf-3310100/sqlite3.c
/bin/sh: python: command not found
make: *** [Release/obj/gen/sqlite-autoconf-3310100/sqlite3.c] Error 127
```

MasterOdin:
For both mac and linux environments, using `node-sqlite3` has mandated first running `npm config set python python3`.

xPaw:
Replacing this python script with a js one should solve the python wonkery: #1570

MasterOdin:
Closing this as I think #1570 is the superior solution where as you say, `tar` is already a transitive dependency so just doing it all in node costs nothing in that regard, and avoids dealing with looking up `python` vs `python3` for the executable.

PR #1550, opened 2022-03-12
Bump SQLite to 3.38.1

eatonphil:
See https://www.sqlite.org/releaselog/3_38_1.html

I had to bump the package.json version to test this out. I can change it to whatever version you want or remove that change.


daniellockyer:
Hey @eatonphil! 👋🏻 We've just updated the bundled version to SQLite v3.38.2 so I'm going to close this - thanks anyway 🙂

PR #1570, opened 2022-04-12
Replaced Python extraction script with JS

xPaw:
node-gyp already depends on tar, so nothing new here.

This should at least reduce some problems related to Python.


daniellockyer:
Thanks @xPaw! 🙂

bpasero:
@daniellockyer should this be a `devDepencency` though?

PR #1615, opened 2022-06-09
chore: swallow non-throwable exceptions

bpasero:
Backport of another VSCode fork related change, authored originally by @deepak1556

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -49,7 +49,7 @@
         "src/node_sqlite3.cc",
         "src/statement.cc"
       ],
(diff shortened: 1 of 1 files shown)


daniellockyer:
@bpasero/@deepak1556 Would you mind explaining the impact of this, or the use-case it solves? 🙂

deepak1556:
We added this as a workaround for [shutdown related crashes](https://github.com/microsoft/vscode/pull/106399#issuecomment-813911734) we were seeing from sqlite3, for context https://github.com/microsoft/vscode/pull/120034#issuecomment-813703991. Addressing https://github.com/TryGhost/node-sqlite3/issues/1455 should help us remove this change.

PR #1628, opened 2022-08-02
Added `package-lock.json` to `.gitignore`

JoelEinbinder:
Running `npm i` in the respository creates a `package-lock.json` file. As this file is not committed, it would be nice to have it in `.gitignore` to avoid the dirty repo state.﻿


daniellockyer:
Thanks! <3

PR #1632, opened 2022-08-30
Fixed importing `sqlite3#verbose` using destructuring syntax

mahdi-farnia:
sqlite3#verbose now returns sqlite3 instead of globalThis object when
bound to undefined (void 0). This enables importing verbose in
typescript using `import { verbose as sqlite3 } from "sqlite3"`. Because
typescript automatically converts import statement to
`sqlite3_1.default.verbose();`.


daniellockyer:
Thank you! 🙂

PR #1633, opened 2022-09-08
Fixed remaining method declarations

alexanderfloh:
This aligns the declarations with the Node-defaults.


daniellockyer:
Nice! Thanks 🙂

PR #1637, opened 2022-09-15
Databese.withPromisses

SrStick:
I create the promise versions of the get, all and run functions in the database class


SrStick:
babel

PR #1654, opened 2022-11-08
Fix rpath linker option when using a custom sqlite

jeromew:
It seems that an old refactoring (~8 years) made a mistake on the way conditions are expressed in gyp. the "conditions" key cannot be set 2 times and the conditions should be all set in the "conditions" array.

The impact of this bug is that when sqlite3 is compiled against a custom sqlite3 source tree, the sqlite3 runtime keeps linking with the system sqlite3 instead of linking to the custom compiled sqlite3 library. In my case that lead to a SIGSEGV upon loading an extension.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -25,8 +25,10 @@
             "libraries": [
                "-l<(sqlite_libname)"
             ],
-            "conditions": [ [ "OS=='linux'", {"libraries+":["-Wl,-rpath=<@(sqlite)/lib"]} ] ],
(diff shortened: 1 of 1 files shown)


jeromew:
For more information,
 - the bug was introduced via https://github.com/TryGhost/node-sqlite3/commit/47f3812c9a5d5bb6efb1d999c1ab5ab7b59e6b8d
 - I originally added the rpath mechanism in https://github.com/TryGhost/node-sqlite3/commit/de0180ccccdd3e2b83c6a7188fd2688c64c70580

daniellockyer:
Thanks!

Issue #1655, opened 2022-11-10
Database.backup() typings?


benallfree:
### Summary

Hi, I noticed there are no TS typings for `Database.backup` and the `Backup` class.

Would you like a PR for that?

https://github.com/TryGhost/node-sqlite3/blob/master/src/backup.h#L17

### Proposed implementation

Add typings and wiki docs for https://github.com/TryGhost/node-sqlite3/blob/master/src/backup.h#L17

daniellockyer:
A PR for this would be great 🙂

ethyaan:
one year later, no changes on this this yet.

benallfree:
@ethyaan Thanks for reminding me, I just made a PR for this.

benallfree:
@daniellockyer Check it out please :)

ethyaan:
@benallfree Thank you, and btw any hands required? just let me know.

viniciusteixeiradias:
Why is this still open?

PR #1656, opened 2022-11-11
#1655 Database.backup() typings

benallfree:
Here's a first cut at #1655. I have some questions:

1. When, exactly, are each of the `callback` params called?
2. Is `Database.backup('foo.db', ()=>console.log('finished'))` a complete example, or do `step()` and `finish` need to be called? 

The docs say

> The `db.backup(filename, [callback])` shorthand is sufficient for making a backup of a database opened by node-sqlite3.

But later in the docs it says that `step(-1)` must be called. 

Then it also says in different places that `finish()` is both required and not required. 

Clearing up these ambiguities would help the dx.

### DIFF
--- a/lib/sqlite3.d.ts
+++ b/lib/sqlite3.d.ts
@@ -259,7 +259,17 @@ export class Backup extends events.EventEmitter {
     retryErrors: number[]
 
     /**
-     * Asynchronously finalize the backup (required).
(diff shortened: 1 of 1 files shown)


daniellockyer:
Closing in favor of https://github.com/TryGhost/node-sqlite3/pull/1726

PR #1664, opened 2022-12-12
Fixed glib compatibility by downgrading to Ubuntu 20

daniellockyer:
fixes https://github.com/TryGhost/node-sqlite3/issues/1663

- Ubuntu 22.04 ships with a newer glibc so the prebuilt binaries end up being linked to a more recent version, removing compatibility with systems that ship with older variants
- this commit fixes that by downgrading the Ubuntu version

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
       matrix:
         os:
           - macos-latest
-          - ubuntu-latest
+          - ubuntu-20.04
           - windows-latest
         host:
           - x64


maxballenger:
Hi, thanks for fixing. It looks like I'm getting hit by this in my project, but I'm not sure exactly how. I am pinned to an older version. This is what my yarn.lock has:

```
sqlite3@^5.0.2:
  version "5.0.8"
  resolved "https://registry.yarnpkg.com/sqlite3/-/sqlite3-5.0.8.tgz#b4b7eab7156debec80866ef492e01165b4688272"
  integrity sha512-f2ACsbSyb2D1qFFcqIXPfFscLtPVOWJr5GmUzYxf4W+0qelu5MWrR+FAQE1d5IUArEltBrzSDxDORG8P/IkqyQ==
  dependencies:
    "@mapbox/node-pre-gyp" "^1.0.0"
    node-addon-api "^4.2.0"
    tar "^6.1.11"
  optionalDependencies:
    node-gyp "8.x"
```

I installed at about 14:35 GMT. Then I got this when trying to run on Ubuntu 18.04:

```
Error: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found
```

I'm thinking this will resolve soon as everything propagates through the pipelines, but I'm confused as to how I ended up getting a glibc 2.33 binary in the first place when I supposedly downloaded version 5.0.8.

Thanks!

maxballenger:
Maybe the issue report is a better place for this comment...

PR #1679, opened 2023-02-12
Expanded Sql for Statement

smith-xyz:
- added macro for expanded sql
- added capability for expanded sql within the context of the callbacks


smith-xyz:
Seeing how this works in more detail, it seemed this was useful to use only within the context of the callbacks. Though, if anything looks off or can be improved, I'm grateful and appreciative of your time and feedback.

PR #1684, opened 2023-02-20
Add openssl_fips to binding.gyp

msotnikov:
Fix this issue
https://github.com/nodejs/node-gyp/issues/2673#issuecomment-1165324060

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -2,6 +2,7 @@
   "includes": [ "deps/common-sqlite.gypi" ],
   "variables": {
       "sqlite%":"internal",
+      "openssl_fips":"",
       "sqlite_libname%":"sqlite3"
   },
   "targets": [


timmypidashev:
Seconding this, would be nice to have latest install without downgrading node-gyp.

daniellockyer:
Can someone explain why this is needed, or at least provide a repro to show the breaking case?

timmypidashev:
> Can someone explain why this is needed, or at least provide a repro to show the breaking case?

Its really an obligatory fix due to the latest node versions not playing nicely with node-gyp. I personally am using an older version(5.1.6) to bypass this issue, as it seems the issue is some breaking feature in 5.1.7. I'm fairly new to node-gyp, so I can't really give an explanation as to why this fixes it. The really fun part is, downgrading node to lts variants or even older releases does not remove this error...

msotnikov:
> Can someone explain why this is needed, or at least provide a repro to show the breaking case?

Because node-gyp has this issues

https://github.com/nodejs/node-gyp/issues/2750
https://github.com/nodejs/node-gyp/issues/2673#issuecomment-1165324060

PR #1686, opened 2023-02-24
Add generic type annotations for Statement and Database get/all/each methods callback rows

stevescruz:
# Description
Small change, but some developers enjoy working with generics over the `any` type when using TypeScript.

Added Generics type annotations for the following methods:
- `Statement.get`
- `Statement.all`
- `Statement.each`
- `Database.get`
- `Database.all`
- `Database.each`

# Backwards Compatibility
This will not break existing TypeScript code. It will still be possible to directly add type annotations to the row/rows.

# Examples:
Take into consideration the following TypeScript code:
```typescript
import sqlite3 from "sqlite3";

interface IUser {
    id: string;
    name: string;
    email: string;
}

const db = new sqlite3.Database("./test.db");
const createTableQuery = `CREATE TABLE IF NOT EXISTS users (
    id TEXT PRIMARY KEY NOT NULL,
    name TEXT NOT NULL,
    email TEXT NOT NULL,
    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
    updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
)`;
```

How type annotations currently works:
**NOTE**: this will still work after the code changes.
```typescript
const query = "SELECT * FROM users";
db.all(query, [], (err, rows: IUser[]) => {
    if (err) {
        console.error(err.message);
    } else {
        console.log(rows[1].email);
    }
});
```

Now you will be able to:
```typescript
const query = "SELECT * FROM users";
db.all<IUser>(query, [], (err, rows) => {
    if (err) {
        console.error(err.message);
    } else {
        console.log(rows[1].email);
    }
});
```

### DIFF
--- a/lib/sqlite3.d.ts
+++ b/lib/sqlite3.d.ts
@@ -80,16 +80,16 @@ export class Statement extends events.EventEmitter {
     run(params: any, callback?: (this: RunResult, err: Error | null) => void): this;
     run(...params: any[]): this;
 
-    get(callback?: (err: Error | null, row?: any) => void): this;
-    get(params: any, callback?: (this: RunResult, err: Error | null, row?: any) => void): this;
+    get<T>(callback?: (err: Error | null, row?: T) => void): this;
+    get<T>(params: any, callback?: (this: RunResult, err: Error | null, row?: T) => void): this;
     get(...params: any[]): this;
 
-    all(callback?: (err: Error | null, rows: any[]) => void): this;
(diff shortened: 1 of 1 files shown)


stevescruz:
@daniellockyer thanks for running the CI pipeline.
We can add the "types" label to this PR.

daniellockyer:
@stevescruz Nice, thanks!

PR #1689, opened 2023-03-03
fix: enable and fix type-casting warnings

rzhao271:
While using [Binskim](https://github.com/microsoft/binskim) to scan native node modules for VS Code, I learned that native node modules have type-casting warnings disabled by default.

This PR re-enables those warnings and fixes them.

The downstream PR, https://github.com/microsoft/vscode-node-sqlite3/pull/32, has some comments on the `sqlite3_bind_*` changes.


rzhao271:
Hi @daniellockyer, I have updated the PR to fix the merge conflicts. I have also adjusted `binding.gyp` so that the build doesn't fail when a type-casting warning occurs.

Issue #834, opened 2017-06-20
move to promises (and async/await)


ORESoftware:
I am not really a fan of the new and shiny, but I am wondering if this library has plans to move to async/await instead of the homegrown flow control currently used. I like the methodology you used (is there a name for it?) but it's hard to mix other async flow control with your library.

For the record - selenium-webdriver did something similar to this library whereby every (async) call was put on a queue - they moved to promises (which will work with async await of course).

You can read about that here:
https://github.com/SeleniumHQ/selenium/wiki/WebDriverJs

and here:
https://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/lib/promise.html

Might be a good excuse to update the lib, or maybe you'd recommend forking it and applying the changes if Mapbox doesn't see a need?

agauniyal:
@ORESoftware this module is rarely updated, even useful PRs are pending for a long time so I don't see if you can get that change easily. Anyways you can use newer module - https://github.com/kriasoft/node-sqlite, which uses this one only underneath the hood.

daniellockyer:
FWIW I'd happily love to see a PR for this - Promises FTW!

PR #1776, opened 2024-04-10
Migrate to Promise-based async implementation

### LINKED ISSUE #1715: streaming support / generator functions
### Summary

is there a way to fetch results of select statement one by one, instead of all at once, by using javascript generator functions?
(those functions that internally use yield statement)

I am aware that .each() function can handle reading rows one by one by using callbacks, 
but callbacks get complicated when used in async/await environment. 
I tried converting ".each()" callbacks to promise and to generator function, but so far this didn't work.

So are generator functions already possible (and how), or could they be added to the core library?

### Proposed implementation

something like 
```js
function* selectEach(query, params = []) ...
```

smoores-dev:
Fixes #1715 and #834

Hi there!

Hopefully this is a welcome request. I've seen @daniellockyer mention a few times that you all would be open to a pull request that moves this library to using a Promise-based async implementation. I'm personally interested in using this library for [Storyteller](https://gitlab.com/smoores/storyteller), so I spent some time this past week working on just that.

## What does this PR do?

This PR replaces the `callback` property on the Baton structs with a `deferred` property, of type `Napi::Promise::Deferred`. Everywhere that was previously calling the callback now either rejects or resolves the `deferred` instead.

The one exception to this is `Statement::Each`, which now returns an `AsyncIterable`. This allows consumers to use the `for await (...)` construct to loop through result rows.

Constructors have been made "private" (in Typescript, at least, and this should be changed in the docs as well). Javascript constructors can't return Promises/be async, so instead, the async initialization code for `Database`, `Statement`, and `Backup` has been split out into their own methods, and new `create` static methods have been added that construct the object and then asynchronously initialize it.

All of the tests have been updated and are passing. The only one that I could use some explanation on is the `async_calls` test; I'm not sure what it was meant to be testing before!

My C++ is rusty at best; I did my best to follow best practices and the patterns laid out in this library, but I may have stumbled. I would absolutely appreciate any and all feedback on my C++ code here!


smoores-dev:
It looks like node-addon-api [somewhat recently added an `engines` spec](https://github.com/nodejs/node-addon-api/commit/afa494ef7f95148a1388813d69628e6367d6f600) to their `package.json`, and it sets a minimum of `^16`, so I had to bump the semver check script to 16. We should probably add our own `engines` spec to this library, given that (and probably even update that semver check script to use that, rather than a hard-coded value). This isn't an actual change in the node versions that this library supports, just making the script more accurately reflect requirements that already existed!

smoores-dev:
Howdy! Just checking in: it looks like the GHA workflow needs to be approved before it can be run. Is this PR something that you all are interested in, generally? No rush; I'm happy to be able to use my own fork while we work on getting it merged, but I wanted to make sure that I'm not pestering you all if you don't want to go in this direction!

smoores-dev:
Hello! @daniellockyer, is this still something that you think you might be interested in? Would it make this change more digestible if it maintained an optional callback interface, in addition to the new Promise one?

huco95:
This would be fantastic to have! Any plans to merge this?

daniellockyer:
Sorry! I'm interested but currently struggling to find time to check it properly and review. Hope to find time soon 🙏🏻

felixnogal:
Hi @daniellockyer, any updates related to this PR?

PR #1830, opened 2025-05-22
Fix fatal error

dnicolson:
Fixes https://github.com/TryGhost/node-sqlite3/issues/1646.

Recreated from https://github.com/TryGhost/node-sqlite3/pull/1693 after the branch was [changed](https://github.com/TryGhost/node-sqlite3/pull/1788#issuecomment-2842110183).


dnicolson:
> @dnicolson could you perhaps add a test case for this issue?

I’m not entirely sure. The error only occurs when there’s a slash in the path, so that part might be worth testing. However, I’m not familiar with that section of the code.

The fix is only at the C++ level like [this](https://github.com/TryGhost/node-sqlite3/pull/1368/files#diff-dc1d7b24ee601a637b09908305be8f69b96d7a9293643ddc88c435923ff648c8L922-L926), to test this I think you would need to use an additional process to capture the exit signal.

PR #1832, opened 2025-05-30
fix: potential null pointer dereference in Blob constructor

kewde:
fixes: https://github.com/TryGhost/node-sqlite3/issues/1827


kewde:
@daniellockyer would you be up to review this?

